Issue #40, opened 2021-05-07
Improve the Login dialog errors for the User


bloodf:
We need to improve the errors on the Login dialog for the user, when there is an wrong password, problematic account or problem with the account creation.

PR #137, opened 2021-08-13
feat: Improve login and register dialogs errors

zfmaster:
## Description
Now responses from createCustomerV2 and generateCustomerToken mutations are handled properly: server errors are still handled as server errors and data errors (wrong password, existing email) messages are proceeded and displayed to FE.

https://devdocs.magento.com/guides/v2.4/graphql/mutations/create-customer-v2.html
https://devdocs.magento.com/guides/v2.4/graphql/mutations/generate-customer-token.html

## Related Issue
https://github.com/vuestorefront/magento2/issues/40

## How Has This Been Tested?

1. Register new customer with available email
2. Register new customer with already taken email
3. Log in with correct customer credentials
4. Log in with wrong customer credentials
5. Make M2 unresponsive and try to log in / register

## Types of changes

## Checklist:

### DIFF
--- a/packages/api-client/src/api/createCustomer/index.ts
+++ b/packages/api-client/src/api/createCustomer/index.ts
@@ -2,7 +2,7 @@ import { FetchResult } from '@apollo/client';
 import {
   CreateCustomerMutation,
   CreateCustomerMutationVariables,
-  CustomerCreateInput, ProductDetailsQuery, ProductDetailsQueryVariables,
+  CustomerCreateInput,
 } from '../../types/GraphQL';
 import createCustomer from './createCustomer';
 import { Context } from '../../types/context';
@@ -20,6 +20,13 @@ export default async (
       fetchPolicy: 'no-cache',
     });
   } catch (error) {
-    throw error.graphQLErrors?.[0].message || error.networkError?.result || error;
+    // For error in data we don't throw 500, because it's not server error
+    if (error.graphQLErrors) {
+      return {
+        errors: error.graphQLErrors,
+        data: null,
+      };
+    }
+    throw error.networkError?.result || error;
   }
 };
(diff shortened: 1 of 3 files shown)


bloodf:
I've found some problems in the implementation, please correct me please if I'm incorrect. 

If the user has a wrong password in the login, or creates a user with a password that doesn't match the required rule, or the email already exists, these errors won't provide to the user the correct problem, they will continue to be in the dark.

zfmaster:
> If the user has a wrong password in the login, or creates a user with a password that doesn't match the required rule, or the email already exists, these errors won't provide to the user the correct problem, they will continue to be in the dark.

Nope, that's not right. It was the scope of the task to make it work, what's the sense of PR if it wouldn't? :smile: 

1. We have 2 parts: composables and api-client. 
2. Before the fix, any non-success response from M2 was treated as error and it was an obstacle in differentiating server error (or network delay) and errors in graphql response from M2. If something was wrong, api-client sent 500 to browser.
3. This PR makes difference: if there are errors in graphql response (what is still according to the schema), we don't send 500 to browser, but send this errors and then they are proceeded in composables.
4. The `logIn` and `register` proceed the information and if result is not what expected the error is thrown, it's catch by useUser of storefront core and placed into `error`.

So, cases which are covered and tested:
1. Server error happens: api-client returns 500 to browser, server error message is displayed
2. Errors returned from M2 graphql: error messages are concatenated with coma (because according to docu M2 could return multiple) and displayed (this is about what was your question)
3. M2 graphql didn't returned error, but also not the token: we display `Customer sign-in error` (according to M2 docu it could never happen, but Magento is not something we have influence to and it's third party service so we can't just rely on documentation but also protect our code from unexpected / undocumented cases). 

I think you were confused with last point, but if you check the code, before we display  `Customer sign-in error` placeholder we first check for errors in M2 response body. 

Here are some screenshots:
Selection_066
Selection_067
 
What we can't get from M2 is `or creates a user with a password that doesn't match the required rule`. Such error simply does not exists: https://devdocs.magento.com/guides/v2.4/graphql/mutations/create-customer-v2.html#errors

Issue #256, opened 2021-12-07
[Bug][My account][Address book]: It's not possible to back from Address details page to Addresses page


Frodigo:
### Expected Behavior

When I open an address form by clicking the "Add new address" button or just edit the existing address and click the "back" browser button, I expect to go back to the Addresses page. 

### Actual Behavior

When I click the "back" browser button, storefront redirects me to another page (which is a previous one before I visited the Addresses details page.

### Possible Solution

Add address details page as a separate route, or add an entry to browser history when the address form appears.

### Steps to reproduce

1. Log in as a user
2. Go to my account
3. Go to the Addresses details page
4. Click the "Add the address" button
5. Click the browser "back" button

### What version of Magento 2 integration are you using?

1.0.0-rc.4.5

### What version of Node.js are you using?

16.13

### What browser (and version) are you using?

Chrome 95

### What operating system (and version) are you using?

macOs Montery 12.0.1

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Frodigo:
The issue has been fixed by @bartoszherba. Here is the corresponding pull request: https://github.com/vuestorefront/magento2/pull/258

Thanks for your help!

@bartoszherba, could you verify if your fix works correctly on our dev environment: https://demo-magento2-canary.europe-west1.gcp.storefrontcloud.io/ ? 

Thanks,
Marcin

bartoszherba:
@Frodigo 
Tested and works as expected.

Frodigo:
Thanks!

PR #258, opened 2021-12-07
fix(theme): 256 customer account - address creation navigation issue

bartoszherba:
Fix issue #256

## Description
Rework customer account navigation to address #256 issue

## Related Issue
Bug #256

## Checklist:

### DIFF
--- a/packages/theme/pages/MyAccount/AddressesDetails.vue
+++ b/packages/theme/pages/MyAccount/AddressesDetails.vue
@@ -115,14 +115,14 @@ export default defineComponent({
     const router = useRouter();
     const route = useRoute();
     const activeAddress = computed(
-      () => userAddresses.value.filter((address) => address?.id == route.value.params.addressId).pop()
+      () => userAddresses.value.filter((address) => address?.id == route.value.query.id).pop()
     );
 
     const isNewAddress = computed(() => !activeAddress.value);
-    const editingAddress = computed(() => !!route.value.params.addressId);
+    const editingAddress = computed(() => !!route.value.query.id);
     const changeAddress = async (address) => {
       const addressId = address?.id || 'new';
-      await router.push({ path: `/my-account/addresses-details/${addressId}` });
+      await router.push({ path: '/my-account/addresses-details', query: { id: addressId } });
     };
 
     const removeAddress = async (address) => {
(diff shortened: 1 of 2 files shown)


bloodf (approved):

packages/theme/routes.js
```diff
@@ -18,7 +18,7 @@ export function getRoutes(themeDir = __dirname) {
   },
   {
     name: 'my-account',
-    path: '/my-account/:pageName?',
+    path: '/my-account/:pageName?/:addressId?',
```
bloodf, line 21: Instead of a parameter, can we use a URL Query?

packages/theme/pages/MyAccount/AddressesDetails.vue
```diff
@@ -110,14 +110,19 @@ export default defineComponent({
...
+    );
 
-    const changeAddress = (address) => {
-      activeAddress.value = address;
-      editingAddress.value = true;
+    const isNewAddress = computed(() => !activeAddress.value);
+    const editingAddress = computed(() => !!route.value.params.addressId);
+    const changeAddress = async (address) => {
+      const addressId = address?.id || 'new';
+      await router.push({ path: `/my-account/addresses-details/${addressId}` });
```
bloodf, line 125: What if we use a URL query to fetch the details, so we don't mess with the routering?

bloodf (changes requested):

packages/theme/pages/MyAccount/AddressesDetails.vue
bartoszherba: The problem with query is that it is not really SEO friendly because of the quotation mark. As we have our custom routing configuration, adding optional parameter is not a big deal, parameter is eventually accessible in router and can be used to conditionally display things or keep address info between pages.

Frodigo: @bloodf, can you elaborate on what you mean by messing with the routing? 

The second question is, what is an advantage of using URL query instead of routing in this case?

bloodf: This page is not even an SEO page, because it's hidden under the MyAccount interface, so it needs to be logged in to be accessed.

The problem with adding new parameters to the routering is that you will add the same parameter that will be valid in all the `my-account` namespace, so if we are on the `my-account/my-orders` the `addressId` will be a valid parameter. 

When using the Query Parameters, you will add the same value without interfering with others' namespace, and in this case, it won't be a problem at all for SEO as this is not found by any SEO robot, as they need to be logged in to use this route.

bartoszherba: Thanks fo reply, I do not know wy I write SEO, what I had in mind was not user friendly url, but in both cases your arguments applies I think. I just updated the PR.

bloodf (approved):

PR #264, opened 2021-12-07
refactor(theme): created new `HTMLContent` component

bloodf:
### 🚀 Features
`theme`
- created new `HTMLContent` component

### 💅 Refactors
`theme`
- removed vue-dompurify-html


packages/theme/components/HTMLContent.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <component
+    :is="tag"
+    v-html="content"
```
bartoszherba, line 4: Maybe I am wrong but this should be computed contentData? Otherwise why do we have that computed const?

Frodigo:
Hi, @bloodf please take a look at @bartoszherba comments, and also rebase the branch with the  develop branch and make sure that build passes. 

Thanks,
Marcin

Frodigo (approved):

PR #265, opened 2021-12-08
fix: fixed build issues and resolved dependencies conflicts

RetroModernDev:
## Description
Fixed build issues and resolved dependencies conflicts. 

## Types of changes

## Checklist:


packages/composables/src/getters/wishlistGetters.ts
```diff
@@ -8,7 +8,7 @@ import {
   Wishlist, WishlistQuery,
 } from '@vue-storefront/magento-api';
 
-type WishlistProduct = WishlistQuery['customer']['wishlists'][0]['items_v2']['items'][0];
+export type WishlistProduct = WishlistQuery['customer']['wishlists'][0]['items_v2']['items'][0] & { variant: any };
```
Mateuszpietrusinski, line 11: can we somehow predict the variant type?

Frodigo: I don't think so because the variant type comes from vsf core. In Magento, there is no variant field on Wishlist, so this is just something like a hack that fixes TS errors and builds. In the real-world, the variant type does not exist in Magento in this place.

bloodf (approved):

PR #314, opened 2021-12-16
refactor(api-client): updated to Apollo 3 interfaces

### LINKED ISSUE #266: Apollo In Memory cache
### What is your question / Please describe your issue

We use apollo-cache-inmemory dependency in api-client but since @apollo/client version 3 that InMemory cache is a part of apollo/client 

Are there any concerns about removing the apollo-cache-inmemory  dependency and using it directly from the @apollo-client? 

Take a look at Apollo docs for reference: 
https://www.apollographql.com/docs/react/caching/cache-configuration/

### What version of Magento 2 Integration are you using?

1.0.0-rc.4.5

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bloodf:
### 💅 Refactors
`api-client`
- updated the interfaces from GraphQL to Apollo 3 compatible
- fixed the Apollo Caching layer to Apollo 3 interfaces
- added locale cookie manipulation to state

`composables`
- added Cookie manipulation on store changing
- added locale cookie manipulation

`theme`
- updated store changing behaviour
- added `nl-NL` flag to icons
- added configs for other currencies on the nuxt-i18n

resolves #266 

⚠️ Breaking Change ⚠️ 
Node.js v16.x is now required to run the integration and the development.


package.json
```diff
@@ -86,6 +86,7 @@
     "vue-eslint-parser": "^8.0.1"
   },
   "engines": {
-    "node": ">=14.x"
+    "node": ">=16.x",
+    "yarn": "1.x"
```
Frodigo, line 90: How about yarn 2, or 3? It will work?

packages/api-client/package.json
```diff
@@ -42,14 +34,16 @@
     "@rollup/plugin-graphql": "^1.1.0",
     "@rollup/plugin-json": "^4.1.0",
     "@types/isomorphic-fetch": "^0.0.35",
-    "apollo-link-schema": "^1.2.5",
     "dotenv": "^10.0.0",
     "graphql-tools": "^8.2.0",
     "jest-transform-graphql": "^2.1.0",
-    "node-fetch": "^2.6.5",
-    "typescript": "^4.5.2",
+    "cross-fetch": "^3.1.4",
```
Frodigo, line 40: Can you explain the reason why we need to change from node-fetch to cross-fetch? Just for better understanding, thanks!

packages/api-client/package.json
```diff
@@ -25,14 +25,6 @@
   },
   "dependencies": {
     "@apollo/client": "^3.5.6",
-    "@vue-storefront/core": "~2.5.2",
```
Frodigo, line 28: @vue-storefront/core is no longer necessary here?

Frodigo: Ok I see it's added as a peer dependency.

packages/api-client/src/helpers/magentoLink/graphQl.ts
```diff
@@ -1,16 +1,20 @@
 /* eslint-disable @typescript-eslint/restrict-template-expressions */
-import ApolloClient from 'apollo-client';
+import {
+  ApolloClient,
+  ApolloLink,
+  HttpLink,
+  InMemoryCache,
+  from,
+} from '@apollo/client/core';
```
Frodigo, line 8: Can be @apollo/client instead of @apollo/client/core?

packages/api-client/src/link.ts
```diff
@@ -1,13 +1,15 @@
-import { createHttpLink } from 'apollo-link-http';
+import {
+  ApolloClient,
+  ApolloLink,
+  InMemoryCache,
+  createHttpLink,
+} from '@apollo/client/core';
```
Frodigo, line 6: The same question as above (about @apollo/client)

Frodigo (changes requested):

package.json
bloodf: No, because how Yarn 2+ handles the node_modules :(

There are still some configurations problems that need to be updated before.

packages/api-client/package.json
bloodf: `cross-fetch` enables the usage of `fetch` everywhere. Preparing for future web IDEs usages :)

packages/api-client/src/helpers/magentoLink/graphQl.ts
bloodf: @Frodigo if we go with `@apollo/client` you are calling the React client. So we need to use the /core

Frodigo (approved):

Issue #247, opened 2021-11-26
[Bug] Broken cart after placing an order as a guest


Fifciu:
### Expected Behavior
After placing an order, cart works correctly.

### Actual Behavior
After creating an order (as a guest), if we want to add a new product to the new cart (without refreshing website - F5) - it won't be reactive. Other actions like sign-in in this circumstances will throw an error.

If we refresh a website with F5 - then everything will work fine - added products will be in cart.

### Possible Solution
Clear cart token after placing an order.

_No response_

### Steps to reproduce
1. Place an order as a guest
1. Without refreshing, go to the product's view and try to add some product - looks like nothing happens
1. If you try to sign-in, it will throw an error because it cannot mergeCarts (looks like something is wrong with new-cart creation process)
1. If you refresh with F5, then everything works fine. It is expected to work like that without a need to refresh website.

### What version of Magento 2 integration are you using?
The newest develop, commit is ba0f10ad80c2fa8c40d281ab45e0e39d09836180

### What version of Node.js are you using?
14.17.6

### What browser (and version) are you using?
Google Chrome 95.0.4638.69 (Official Build) (64-bit)

### What operating system (and version) are you using?
Ubuntu 20.04

### Code of Conduct
* [x] I agree to follow this project's Code of Conduct

PR #386, opened 2022-01-07
fix: broken cart after placing order as guest

KevinGorjan:
Fixed when a user wants to place an order as a guest.

## Description
Fix for #247 
The problem was that the guest email address was not set correctly. 
Also fixed when passing the "neighborhood" and "extra" param when setting the billing address and shipping address.
If the params are empty, they are filtered out
Updated tests

## Motivation and Context
Now a guest can place an order

## Types of changes

## Checklist:

### DIFF
--- a/packages/composables/src/composables/useShipping/index.ts
+++ b/packages/composables/src/composables/useShipping/index.ts
@@ -49,7 +49,7 @@ const factoryParams: UseShippingParams<any, any> = {
       : ({
         address: {
           ...address,
-          street: [address.street, apartment, neighborhood, extra].filter(Boolean), // remove unused attributes
+          street: [address.street, apartment, neighborhood, extra].filter(Boolean),
         },
       });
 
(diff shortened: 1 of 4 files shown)


packages/composables/src/composables/useShipping/index.ts
```diff
@@ -49,7 +49,7 @@ const factoryParams: UseShippingParams<any, any> = {
       : ({
         address: {
           ...address,
-          street: [address.street, apartment, neighborhood, extra],
+          street: [address.street, apartment, neighborhood, extra].filter(Boolean), // remove unused attributes
```
bloodf, line 52:
```suggestion
          street: [address.street, apartment, neighborhood, extra].filter(Boolean),
```

bloodf (approved):
Just removing an unnecessary comment :)

Frodigo (approved):
Good job, @KevinGorjan !

PR #408, opened 2022-01-13
feat(middleware): added i18n route middleware

### LINKED ISSUE #378: [Bug]: Change store / locale
### Expected Behavior
_No response_

### Actual Behavior
When i change a store / locale from a language switcher it works good.
But when i change a store / locale in the url it changes after a second reload page and doesn't logout and clear cart.

### Possible Solution
_No response_

### Steps to reproduce
_No response_

### What version of Magento 2 integration are you using?
2.4.3

### What version of Node.js are you using?
v16.12.0

### What browser (and version) are you using?
Chrome Version 96.0.4664.93 (Official Build) (x86_64)

### What operating system (and version) are you using?
macOS Version 12.0.1 (21A559)

### Relevant log output
_No response_

### Code of Conduct
* [x] I agree to follow this project's Code of Conduct

Frodigo:
## Description
Added i18n middleware to handle stores changing by URL

## Related Issue
closes #378

## How Has This Been Tested?
- Added unit tests
- tested manually 

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.


packages/composables/nuxt/plugin.js
```diff
@@ -76,6 +77,18 @@ export default integrationPlugin(({ app, res, req, integration }) => {
     setCookie(localeCookieName, id);
   };
 
+  const getCountry = () => getCookies(countryCookieName);
+
+  const setCountry = (id) => {
+    if (!id) {
```
bartoszherba, line 83: !id ? removeCookie(countryCookieName) : setCookie(countryCookieName, id)

packages/composables/nuxt/plugin.js
```diff
@@ -76,6 +77,18 @@ export default integrationPlugin(({ app, res, req, integration }) => {
...
+      // eslint-disable-next-line no-param-reassign
+      removeCookie(countryCookieName);
+      return;
+    }
+
+    setCookie(countryCookieName, id);
```
bartoszherba, line 89:
```suggestion
      // eslint-disable-next-line no-param-reassign
      !id ? removeCookie(countryCookieName) : setCookie(countryCookieName, id);
```

packages/theme/middleware/i18n.js
```diff
@@ -0,0 +1,88 @@
...
+    return;
+  }
+
+  const i18nLocaleObject = findLocaleBasedOnStoreCode(currentStoreCode, i18n.locales);
+
+  if (!i18nLocaleObject) {
+    await setDefaultLocale(i18n);
+
+    return;
+  }
```
bartoszherba, line 74:
```suggestion
  if (!currentStoreCode || !findLocaleBasedOnStoreCode(currentStoreCode, i18n.locales)) {
    await setDefaultLocale(i18n);

    return;
  }
```

packages/theme/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "nuxt build -m",
     "build:analyze": "nuxt build -a -m",
-    "dev": "nuxt",
+    "dev": "nuxt --inspect",
```
bartoszherba, line 13: node --inspect flag affects performance, maybe we should add something like "dev-debug": "nuxt --inspect" here and in the root `yarn dev-debug:theme` or even just `yarn dev-debug`, both runs themes dev-debug cmd.

packages/theme/middleware/i18n.js
```diff
@@ -0,0 +1,88 @@
+import cookieNames from '~/enums/cookieNameEnum';
```
bartoszherba, line 1: In general, I think that this script is not well designed.
It exposes a lot of public methods which either should be private or should belong to a dedicated cookie management class. In the current situation we have middleware which, if we want to utilize its methods, must be imported in a file we want to use it. IMO middleware, as configured service run in a particular lifecycle, mustn't be imported and used anywhere but only as a configured middleware. We are mixing responsibilities which leads to bad development experience and coupled architecture. What do you think if we extract cookie management class to handle cookie manipulation class/script (name it) and let middleware delegate tasks.

Exqample, I do not want to import middleware to setDefaultLocale, if so - method should be in a separate class and should not be exported in middleware.

packages/theme/middleware/i18n.js
```diff
@@ -0,0 +1,88 @@
...
+    cookie += `vsf-customer=${customerTokenCookie}; `;
+  }
+
+  const cartIdCookie = apiState.getCartId();
+
+  if (cartIdCookie) {
+    cookie += `vsf-cart=${cartIdCookie} `;
+  }
+
+  console.log(cookie)
```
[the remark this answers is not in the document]
Frodigo: removed

packages/composables/nuxt/plugin.js
Frodigo: Done

packages/theme/middleware/i18n.js
Frodigo: Done

packages/theme/package.json
Frodigo: good catch, added `dev-debug` script

packages/theme/middleware/i18n.js
Frodigo: Refactored.

Frodigo:
@bartoszherba thanks for your comments. All issues were resolved, and some code refactored. 

Please, re-check. 

Thanks,
Marcin

Issue #410, opened 2022-01-13
[Bug]: Loosing data when moving between steps on the checkout


bartoszherba:
### Expected Behavior
Whenever I switch my checkout step to the one I have already visited and validated I expect to have the same already data populated.

### Actual Behavior
When I switch between tabs I loose my data (at least on the frontend visible part)

### Steps to reproduce
1. Add something to the cart and visit the checkout
1. Populate data in the first form
1. Proceed to the second step
1. Go back to the previous step by clicking on step's tab bar link
1. Observe that data you added before are missing

### What version of Magento 2 integration are you using?
2.4.x

### What version of Node.js are you using?
16.13.1

### What browser (and version) are you using?
Version 96.0.4664.110 (Official Build) (arm64)

### What operating system (and version) are you using?
macOS Monterey 12.1

### Code of Conduct
* [x] I agree to follow this project's Code of Conduct

┆Attachments: <a href="https://vsf.atlassian.net/rest/api/2/attachment/content/11065">image-20220124-125759.png</a>

Issue #409, opened 2022-01-13
[Bug]: You can go to any checkout step without previous validation


bartoszherba:
### Expected Behavior
Whenever I visit checkout I can go through or go to a particular step only if previous step is completed and validated.

### Actual Behavior
Once I land on checkout I can go to any step without validation of previous steps, eg. I can instantly go to the last step and try to place an order.

### Possible Solution
Checkout navigation must follow rules:

* I can go to the next step only if the previous step is already completed and validated
* I can revisit any previous step at any time
* If I revisit one of the previous steps and change any data I will have to revalidate all next steps again (as previous data might have an impact on further calculation)

### Steps to reproduce
Simply go to the checkout and instantly switch to the last step. You can even try to place an order without previously validated steps.

### What version of Magento 2 integration are you using?
2.4.x

### What version of Node.js are you using?
16.13.1

### What browser (and version) are you using?
Version 96.0.4664.110 (Official Build) (arm64)

### What operating system (and version) are you using?
macOS Monterey 12.1

### Code of Conduct
* [x] I agree to follow this project's Code of Conduct

┆Attachments: <a href="https://vsf.atlassian.net/rest/api/2/attachment/content/11089">image-20220125-133016.png</a>

Frodigo:
@igorwojciechowski thanks for catching this. We have separate ticket for that: https://github.com/vuestorefront/magento2/issues/492

PR #440, opened 2022-01-18
fix: loosing data when moving between steps on the checkout

KevinGorjan:
This ticket fixes 
* #410

## Description
The user can't navigate to any next page anymore, this can only be done when the form from the current step correctly is filled in.
If the user goes directly to the page by URL, the application will check of the previous step is valid, else the user will be redirected to the user-account page

## Related Issue
* #409  

## Types of changes

## Checklist:


packages/theme/pages/Checkout/UserAccount.vue
```diff
@@ -251,14 +253,23 @@ export default defineComponent({
...
         form.value.firstname = user.value.firstname;
         form.value.lastname = user.value.lastname;
         form.value.email = user.value.email;
       }
     });
 
+    onMounted(async () => {
+      const checkout = await getItem('checkout');
+      if (checkout && checkout['user-account']) {
+        const data = checkout['user-account'];
```
Frodigo, line 266: I didn't see the logic for filling data in other steps. It's possible to do something like this in other steps as well?

Frodigo (changes requested):

packages/theme/pages/Checkout/UserAccount.vue
KevinGorjan: On the shipping and billing step, the form gets the data when fetching the cart.

image

That's why I don't store it in the localstorage.

packages/theme/pages/Checkout/Payment.vue
```diff
@@ -209,11 +211,20 @@ export default defineComponent({
       await load();
     });
 
+    onMounted(async () => {
+      const checkout = await getItem('checkout');
+      if (!checkout || !checkout.billing) {
```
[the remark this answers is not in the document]
KevinGorjan: Good idea, I'll add this ;-)

KevinGorjan:
@Frodigo I added tests for the asyncLocalStorage and the steps-helper for the checkout.

I didn't add tests in the pages because this will be done in a separate ticket

Frodigo (approved):

PR #441, opened 2022-01-19
fix: handle currency switch issue

bartoszherba:
## Description
Fix for a currency switcher and currency formatting

## Related Issue
#436 

## Motivation and Context
Even though the currency was properly switched and prices were recalculated, i18n was not formatting prices properly. It is because i18n can handle only one currency for a given locale. Therefore I introduced a new format currency plugin that will overcome that limitation.

## How Has This Been Tested?
switch currency for multiple stores and observe if currency symbol and rates are converted properly

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.


packages/theme/plugins/fcPlugin.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export default ({ app }, inject) => {
+  inject('fc', (value: number | string, locale?: string, options = {}): string => {
+    // eslint-disable-next-line no-param-reassign
+    locale = locale || app.$cookies.get(cookieNames.localeCookieName);
+    // eslint-disable-next-line no-param-reassign
+    options = { currency: app.$cookies.get(cookieNames.currencyCookieName), ...options };
+
+    return formatCurrency(value, locale, options);
```
Frodigo, line 19: @bartoszherba  Just a question: does it works when the locale is for example `default`? In locale cookies, we keep the Magento store code. Wi thought that in this place you need to pass ISO locale code which is available on locale config object.

bloodf: Please check the Currency Switch implementation on handling with the locale and ISO formats :)

bartoszherba: @Frodigo Well ok, I changed that, I did not know that we have arbitrary locale names. On the other hand ppl might not expect to get 'default' or a store code from a locale cookie especially since we have vsf-store cookie which is, as I understand now, exactly the same.

Frodigo (approved):

Issue #406, opened 2022-01-13
[Bug]: Once the customer log out his cart is not cleared


bartoszherba:
### Expected Behavior
When I log out from my account my session and cookies are cleared, all session-related data are also cleared on the frontend (especially cart data, as this is an obviously observable issue)

### Actual Behavior
When I log out from my account my session and cookies are **not** cleared, first obvious and observable issue is related to a cart.

### Steps to reproduce
1. Log In as a customer
1. Add product to a cart unless you have any
1. Log out
1. Observe that cart is not cleared

### What version of Magento 2 integration are you using?
2.4.x

### What version of Node.js are you using?
16.13.1

### What browser (and version) are you using?
Version 96.0.4664.110 (Official Build) (arm64)

### What operating system (and version) are you using?
macOS Monterey 12.1

### Code of Conduct
* [x] I agree to follow this project's Code of Conduct

PR #458, opened 2022-01-20
fix(composable): fix clearing cart on customer logout

bartoszherba:
## Description
Clearing cart data on customer logout.

## Related Issue
#406 

## Motivation and Context
Bugfix

## How Has This Been Tested?

1. Log in as a customer
2. Add something to cart
3. Log out
4. Observe that your cart is cleared

## Types of changes

## Checklist:

### DIFF
--- a/packages/composables/src/composables/useUser/index.ts
+++ b/packages/composables/src/composables/useUser/index.ts
@@ -46,7 +46,7 @@ CustomerCreateInput
 
     apiState.setCustomerToken(null);
     apiState.setCartId(null);
-    // context.cart.setCart(null);
+    context.cart.setCart(null);
   },
   updateUser: async (context: Context, params) => {
     Logger.debug('[Magento] Update user information', { params });


packages/composables/src/composables/useUser/index.ts
```diff
@@ -18,7 +18,7 @@ CustomerCreateInput
       cart: useCart(),
     };
   },
-  load: async (context: Context, params) => {
+  load: async (context: Context) => {
```
Frodigo, line 21: I am wondering why params are not used here? Probably it's a bug and custom query won't work here. 
Screenshot 2022-01-20 at 12 34 57

customer method in API should receive custom query: `customer(customQuery?: CustomQuery): Promise<ApolloQueryResult<CustomerQuery>>;`

Can you add logic responsible for custom query here?

Frodigo (changes requested):

packages/composables/src/composables/useUser/index.ts
bartoszherba: I do not understand now how it works, there is a lot of places where API might expect custom query but do not get any, as it is here. I must understand this problem but this is a completely different thing and I do not want to block this tack. I removed that change and we can do it in a separate task.

Frodigo (approved):

PR #541, opened 2022-02-02
refactor(theme): performance improvements

Frodigo:
## Description
- Removed chunks configuration that causes performance issues relates to prefetching js scripts

## How Has This Been Tested?
Tested on local, needs to be confirmed on the cloud as well.

### Without redis cache:
Screenshot 2022-02-02 at 11 50 13

### With Redis cache: 
Screenshot 2022-02-02 at 12 00 22

## Types of changes

## Checklist:


bartoszherba:
@Frodigo 
I think that this is a bad idea. 

- you cannot predict how long does it take for a browser to become idle, I faced that problem before for smaller components that were non-reactive for a too long time though the web was rendered.
- you introduced the component which is actually not a component, just a hacky wrapper for some logic
- according to lifecycles onMounted should be exactly what you want https://nuxtjs.org/docs/concepts/nuxt-lifecycle/ without the introduction of an awkward component
- hydration occurs before mounting and I think that hydration itself is not a problem in this case
- IMO this is ad hoc solution, we should ask why it takes so long to load the config in the first place; maybe we should never hit the server, but regardless, config is cached on magento side, it should not block rendering 
- what was the previous score and what was the sampling to measure the difference?

Frodigo:
@bartoszherba, good point. I removed that MagentoConfiguration component, and it looks like chunks configuration caused the problems.

bartoszherba (approved):

PR #562, opened 2022-02-03
chore: fixed some error reported by eslint

Frodigo:
## Description
I just ran `eslint --fix` command....

## Types of changes

## Checklist:


packages/theme/types/index.ts
```diff
@@ -0,0 +1,3 @@
+interface UseContextReturn {
+  $recaptcha: any
```
bartoszherba, line 2: I am not sure what is this? Using any is a generally bad practice as it is meaningless unless there is a very good reason to use any. Also, I do not see where do you use this interface?

bartoszherba, line 2: I am not sure what is this? Using any is a generally bad practice as it is meaningless unless there is a very good reason to use any. Also, I do not see where do you use this interface?

bartoszherba, line 2: I am not sure what is this? Using any is a generally bad practice as it is meaningless unless there is a very good reason to use any. Also, I do not see where do you use this interface?

bartoszherba (dismissed):
@Frodigo One small comment, probably redundant file was included.

packages/theme/types/index.ts
Frodigo: @bartoszherba this is a mistake, removed.

bartoszherba (approved):

PR #571, opened 2022-02-07
refactor: refactored useExternalCheckout composable

Frodigo:
## Description
- mark composable useExternalCheckout as deprecated
- mark factory useExternalCheckout as deprecated
- move composable to theme module
- changed initializeCheckout method binding in CartSidebar

## Types of changes

## Checklist:

### DIFF
--- a/packages/composables/src/composables/useExternalCheckout/index.ts
+++ b/packages/composables/src/composables/useExternalCheckout/index.ts
@@ -3,6 +3,11 @@ import { useExternalCheckoutFactory, UseExternalCheckoutFactoryParams } from '..
 import useCart from '../useCart';
 // import useConfig from '../useConfig';
 
+/**
+ * @deprecated since version <version?>
+ *
+ * @see <add docs link>
+ */
 const factoryParams: UseExternalCheckoutFactoryParams = {
   provide() {
     return {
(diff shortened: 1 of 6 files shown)


packages/composables/src/index.ts
```diff
@@ -17,7 +17,6 @@ export { default as useCountrySearch } from './composables/useCountrySearch';
 export { default as useCurrency } from './composables/useCurrency';
 export { default as useCustomMutation } from './composables/useCustomMutation';
 export { default as useCustomQuery } from './composables/useCustomQuery';
-export { default as useExternalCheckout } from './composables/useExternalCheckout';
```
bartoszherba, line 20: Maybe mark as deprecated instead, eg i use
```
/**
 * @deprecated since version <version?>
 *
 * @see <add docs link>
 */
```
So we can remove it later at some point to not break other's apps. If someone is using this composable for example

bartoszherba (changes requested):

packages/theme/composables/useExternalCheckout/useExternalCheckout.ts
```diff
@@ -0,0 +1,57 @@
...
+    } finally {
+      loading.value = false;
+    }
+
+    return null;
+  };
+
+  return {
+    initializeCheckout,
+    loading: readonly(loading),
```
bartoszherba, line 52: Readonly unwrap REF so you will be not able to do something like loading.value in code, is this what you want?

Frodigo: I see, no

Frodigo:
@bartoszherba this is ready for re-review

bartoszherba (approved):

PR #572, opened 2022-02-07
refactor(theme): refactor useConfig composable

bartoszherba:
## Description
    - mark composable useConfig as deprecated
    - move composable to theme module
    - rework composable to get rid of core dependency and to utilize Pinia store

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/composables/useConfig/index.ts
+++ b/packages/theme/composables/useConfig/index.ts
@@ -1,0 +1,35 @@
+import {
+  computed, ref, useContext,
+} from '@nuxtjs/composition-api';
+import { Logger } from '@vue-storefront/core';
+import { useConfigStore } from '~/stores/config';
+import { UseConfig } from '~/composables/useConfig/useConfig';
+
+const useConfig = (): UseConfig => {
+  const { app } = useContext();
+  const loading = ref(false);
+  const configStore = useConfigStore();
+
+  const loadConfig = async () => {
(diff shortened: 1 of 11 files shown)


packages/theme/composables/useConfig/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    } catch (err) {
+      Logger.debug(err);
+    }
+
+    loading.value = false;
+  };
+
+  return {
+    config: computed(() => configStore.value),
+    loading,
```
Frodigo, line 30: what do you think about wrapping this by `readonly` from composition API?

packages/theme/composables/useConfig/index.ts
```diff
@@ -0,0 +1,35 @@
...
+      });
+    } catch (err) {
+      Logger.debug(err);
+    }
+
+    loading.value = false;
+  };
+
+  return {
+    config: computed(() => configStore.value),
```
Frodigo, line 29: Does it need to be computed?

Frodigo (changes requested):

packages/theme/composables/useConfig/index.ts
bartoszherba: Otherwise, it will break the interface and I will have to update types and logic in other places. For now let it be.

PR #585, opened 2022-02-10
refactor: refactored useCategory composable

Frodigo:
## Description
- refactored useCategory composable
- added useNavigation composable
- added navigation pinia store

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/components/AppHeader.vue
+++ b/packages/theme/components/AppHeader.vue
@@ -252,7 +252,7 @@ export default defineComponent({
       return data?.categories?.items;
     });
 
-    const categoryTree = computed(() => categoryGetters.getCategoryTree(categoryList.value?.[0])?.items.filter((c) => c.count > 0));
+    const categoryTree = computed(() => categoryGetters.getCategoryTree(categoryList?.value?.[0])?.items.filter((c) => c.count > 0));
 
     const handleAccountClick = async () => {
       if (isAuthenticated.value) {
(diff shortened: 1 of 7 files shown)


packages/theme/composables/useNavigation/index.ts
```diff
@@ -0,0 +1,28 @@
+import { useAsync } from '@nuxtjs/composition-api';
```
bartoszherba, line 1: I REALLY do not like this composable, especially reflecting NAVIGATION in the store. You did it because you are using mobile and desktop devices but we should not make an API call or render a mobile menu if we are on desktop! To my taste, this is the store overuse. Even then, different components might want to have different navigation layouts and data, you assumed that mobile and desktop are the same, and there is no way to parametrize or reload data. Even if we agree on that store, I hope we will not, then there is no way to pull different data for mobile or desktop or footer or any possible component. Please remove this composable and simply do not render mobile menu for mobile, or rethink how it should work.

bartoszherba (changes requested):
@Frodigo Please, check my comment about NavigationComposable

packages/theme/composables/useNavigation/index.ts
Frodigo: @bartoszherba regarding our discussion on Slack, I changed my behavior.

bartoszherba (approved):

PR #677, opened 2022-03-01
feat: migrate to esbuild

filipsobol:


packages/composables/package.json
```diff
@@ -22,7 +28,7 @@
     "update:update": "ncu -u"
   },
   "dependencies": {
-    "@vue-storefront/core": "~2.5.4",
+    "@vue-storefront/core": "link:/home/fsobol/Projects/Storefront/packages/core",
```
Frodigo, line 31: Looks like a mistake, please undo this change.

packages/theme/package.json
```diff
@@ -34,19 +34,18 @@
     "@nuxtjs/pwa": "^3.3.5",
     "@nuxtjs/style-resources": "^1.2.1",
     "@storefront-ui/vue": "^0.11.5",
-    "@vue-storefront/cache": "^2.5.4",
-    "@vue-storefront/core": "~2.5.4",
+    "@vue-storefront/cache": "link:/home/fsobol/Projects/Storefront/packages/cache",
```
Frodigo, line 37: Same in this file

Frodigo (changes requested):

Frodigo (dismissed):

packages/api-client/package.json
```diff
@@ -7,13 +7,20 @@
...
   "types": "lib/index.d.ts",
   "scripts": {
-    "build": "rimraf lib server && rollup -c",
-    "dev": "rimraf lib server && rollup -c -w",
+    "build": "yarn build:clear && yarn build:package && yarn build:server && yarn build:types",
+    "build:clear": "rimraf lib",
+    "build:package": "node ../../scripts/build --file=src/index.ts",
+    "build:server": "node ../../scripts/build --file=src/index.server.ts",
+    "build:types": "tsc --emitDeclarationOnly",
+    "dev": "yarn dev:package & yarn dev:server",
```
sethidden, line 23: Wondering if this works on Windows? Asking because of the I think UNIX-only "&" syntax
---
Won't this also cause a zombie node process if you CTRL+C in the terminal (only the process in the foreground will be killed, no)? You can easily check this by doing something like `npm install & npm install` - if you ctrl C, you'll still see output from the 2nd process

Guess you could use https://www.npmjs.com/package/npm-run-all or some yarn builtin if it exists (I'm an NPM man myself)

filipsobol: That's an excellent point! I fixed it in https://github.com/vuestorefront/magento2/pull/677/commits/846b00a400830ee5f081ee7bad15a29e93b600d0

sethidden:
LGTM, I'll talk to Marcin tomorrow about this

sethidden (approved):

Frodigo (approved):

PR #678, opened 2022-03-02
build: refactored precommit hooks execution

Frodigo:
## Description
Fixed recommit hooks execution. 
Since git doesn't support running lint-staged in parallel with Lerna we need to run `precommit` commands synchronously.

## Types of changes

## Checklist:

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,12 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx lerna run --parallel precommit
+rootDir=$(pwd)
+
+# lint-staged cannot run in parallel cuz git, so we manually loop over
+# each package and run the precommit command
+for package in $(ls packages);
+do
+    cd "$rootDir/packages/$package"
+    yarn precommit
+done


.husky/pre-commit
```diff
@@ -1,4 +1,14 @@
+# .husky/precommit
+
```
sethidden, line 2: Are these lines necessary? In general the shebang #!/bin/sh should be on the first line in the file

.husky/pre-commit
```diff
@@ -1,4 +1,14 @@
...
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx lerna run --parallel precommit
+rootDir=$(pwd)
+
+# lint-staged cannot run in parallel cuz git, so we manually loop over
+# each package and run the precommit command
+for package in $(ls packages);
```
sethidden, line 10: (moving convo from slack to here)
Worried about the Linux commands `ls` etc. that are not available in Windows's cmd.exe
Theoretically if a contributor has Git for Windows installed they should have some basic Linux commands like `sh` or `ls`, but not sure if they're added to PATH

sethidden:
In general, I copied the script to my local repo, ran `yarn` (to refresh the precommit hooks) and it works for me

.husky/pre-commit
```diff
@@ -1,4 +1,14 @@
+# .husky/precommit
+
 #!/bin/sh
```
sethidden, line 3:
```suggestion
#!/bin/sh
```
Are these lines necessary? In general the shebang #!/bin/sh should be on the first line in the file

sethidden: I checked on a Windows machine and the shebang needs to be on the first line or else you'll get a "cannot spawn" error, so the change I suggested is necessary

.husky/pre-commit
sethidden: This works on Windows if you have Git for Windows and the necessary "ls" etc. utils in your PATH

sethidden (approved):

PR #680, opened 2022-03-02
fix(theme): category quantity input value omitted when adding product to cart

ghost:
## Description
Fixed the reported issue where adding items from the list view would add them to cart always with quantity 1, even if the user provided something else eg. 1337

Since it's a change to business logic I really wanted to write some tests for this, but there's a lot of boilerplate code to even start (which would turn this 5 min task into a 6 hour one)

Also fixed some linting errors as husky started bothering me because I touched the file

## Related Issue
#662 

## How Has This Been Tested?
Tested manually the two scenarios:
1. Add many products to cart (e.g. 1337)
2. Add one (to check if the value "null" is not emitted or something when there's just 1 - there's not)

N/A

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/pages/Category.vue
+++ b/packages/theme/pages/Category.vue
@@ -123,7 +123,7 @@
                 :title="getName(product)"
                 :wishlist-icon="isAuthenticated ? '' : ''"
                 class="products__product-card-horizontal"
-                @click:add-to-cart="addItemToCart({ product, quantity: 1 })"
+                @click:add-to-cart="addItemToCart({ product, quantity: $event })"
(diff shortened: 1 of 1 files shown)


packages/theme/components/WishlistSidebar.vue
```diff
@@ -250,7 +250,9 @@ export default defineComponent({
 .empty-wishlist {
   display: flex;
   flex: 1;
+  align-items: center;
```
sethidden, line 253: whoops this shouldn't be here

bartoszherba (approved):

PR #687, opened 2022-03-04
feat(theme): refactor category tree

bartoszherba:
## Description

- category tree is split to separate components
- composables are not used, use the new useApi hook to query for the data
- create dedicated Modules/Catalog space
- remove all dependencies to composables and api-client
- move types to Catalog/Category module
- add module scoped routes resolver

With this PR I want also to introduce a more systematic and meaningful directories setup. I introduce a Modules space where we can reflect Magento modules structure and keep related parts of the application.

Also, I propose components namespacing convention and a rule to follow - to keep code close components and close to modules. It is not against keeping some generic functions or plugins but:
- if something is a helper it should be moved closer to the module or component and to be renamed to reflect an idea behind it
- if plugins are not generic but closely related to module/component keep it close
- as a rule of thumb, keep code as close as possible unless it makes sense to have it in the upper scope

....<I will add more here once it all is done> 
## Types of changes

## Checklist:


packages/theme/Modules/Catalog/Category/Sidebar/command/loadCategoriesCommand.ts
```diff
@@ -0,0 +1,22 @@
+import categoryListQuery from './categoryList.gql';
+import { CategoryInterface } from '~/Modules/Catalog/Category/types';
+import useApi from '~/composables/useApi';
+
+export const loadCategoriesCommand = {
+  execute: async (customQuery: string = categoryListQuery): Promise<CategoryInterface[]> => {
```
Frodigo, line 6: this is great, thank you🚀

Frodigo (approved):

Frodigo (approved):

PR #681, opened 2022-03-02
feat(theme): add basic skeleton loader component

bartoszherba:
## Description
I added a new basic SkeletonLoader component.
This is the first iteration, we must decompose category further.

## Related Issue
-

## Motivation and Context
-

## How Has This Been Tested?
-

## Screenshots (if appropriate):
-vue-storefront-magento-theme

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #690, opened 2022-03-08
fix(theme): fix infinite request loop when adding single quantity product

### LINKED ISSUE #686: [Category] Adding product to cart triggers infinite request loop
### Expected result
Product is added to cart

### Actual result
Clicking add to cart button triggers infinite request loop

ghost:
Fix bug introduced in #681 where quantity would not be sent (undefined) - overlooked the extra $event back when I was reviewing that one (should be in just the horizontal card, not in both)

Fixes #686 

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/pages/Category.vue
+++ b/packages/theme/pages/Category.vue
@@ -78,7 +78,7 @@
               :wishlist-icon="isAuthenticated ? 'heart' : ''"
               class="products__product-card"
               @click:wishlist="addItemToWishlist(product)"
-              @click:add-to-cart="addItemToCart({ product, quantity: $event })"
+              @click:add-to-cart="addItemToCart({ product, quantity: 1 })"
             >
               <template #image="imageSlotProps">
                 <SfButton


packages/theme/pages/Category.vue
```diff
@@ -78,7 +78,7 @@
               :wishlist-icon="isAuthenticated ? 'heart' : ''"
               class="products__product-card"
               @click:wishlist="addItemToWishlist(product)"
-              @click:add-to-cart="addItemToCart({ product, quantity: $event })"
+              @click:add-to-cart="addItemToCart({ product, quantity: 1 })"
```
Frodigo, line 81: but what in case if the user wants to add more than 1 item to the cart?

RetroModernDev (changes requested):

packages/theme/pages/Category.vue
sethidden: This is for the simple "add to cart" card 
obraz

The SfCardHorizontal below has $event
obraz

Frodigo: I see, thanks!

RetroModernDev (approved):

PR #730, opened 2022-03-14
chore(theme): refactor getters from composables to theme

youanden:
## Related Issue

https://vsf.atlassian.net/browse/M2-261

Work in progress

## Types of changes

## Checklist:


packages/theme/getters/addressGetter.ts
```diff
@@ -1,5 +1,4 @@
-import { Countries, Country } from '@vue-storefront/magento-api';
-import { AddressGetter } from '../types';
+import { Countries, Country, AddressGetter } from '@vue-storefront/magento';
```
sethidden, line 1: @Frodigo is taking types from /magento better than /magento-api in this case? We have a separate task for migrating the types into the theme, but just wondering though

packages/theme/getters/categoryGetters.ts
```diff
@@ -1,6 +1,5 @@
 import { CategoryGetters, AgnosticCategoryTree, AgnosticBreadcrumb } from '@vue-storefront/core';
-import { Category } from '@vue-storefront/magento-api';
-import { buildCategoryTree } from '../helpers/buildCategoryTree';
+import { Category, buildCategoryTree } from '@vue-storefront/magento';
```
sethidden, line 2:
```suggestion
import { Category } from '@vue-storefront/magento';
```
We already have our own buildCategoryTree instance within the `theme` package. Can you please use it instead? They're duplicated for now but I think the package version will be removed soon

packages/theme/getters/productGetters.ts
```diff
@@ -6,13 +6,11 @@ import {
   ProductGetters as ProductGettersBase,
 } from '@vue-storefront/core';
 import {
-  BundleProduct,
+  htmlDecode, BundleProduct,
```
sethidden, line 9:
```suggestion
  BundleProduct,
```
Same as in previous comment - please use the `theme` version (import {htmlDecode} from '../helpers/htmlDecoder')

Frodigo: Hmm, I can't find htmlDecoder in theme.

sethidden: Right, misread on my side

Frodigo (changes requested):
@youanden make sure that build pass.

Frodigo (approved):

Issue #735, opened 2022-03-15
[Feature]: improve performance of calls between Magento and VSF


marc-arnoult:
### How the project can be improved?

With a Headless Framework like **VSF**, we are making most of our calls to the e-commerce backend from NodeJS.
The **VSF** backend is more or less a _Proxy_, so in order to improve the performance we can reuse the existing connection between **VSF** and **Magento**.

### What are the acceptance criteria?

- [ ]

### Additional information

I found that every call made by **VSF** was slower than an API Client like Postman for example. 
After a few research i found that we are losing time to make a connection ( HTTPS round trip and so on... ).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Frodigo:
This will be released in the upcoming 1.0.0-rc.7 release.

PR #737, opened 2022-03-16
feat(api-client): improve performance of node HTTP calls

marc-arnoult:
## Related Issue
Resolve https://github.com/vuestorefront/magento2/issues/735

## Motivation and Context
I found that every call made by VSF was slower than an API Client like Postman for example. After a few research i found that we are losing time to make a connection ( HTTPS round trip and so on... ).

## How Has This Been Tested?
- Home/Product/Categories/MyAccount pages have been tested.
- Add to cart, remove, update.
- Login/Logout

Capture d’écran 2022-03-15 à 08 52 33
Capture d’écran 2022-03-15 à 08 49 19

:warning: this is on my local Machine, on production we can expect 10 to 30% improvement i think.

## Checklist:

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -35,7 +35,8 @@
     "graphql": "^16.3.0",
     "graphql-tag": "^2.12.6",
     "isomorphic-fetch": "^3.0.0",
-    "dotenv": "^15.0.0"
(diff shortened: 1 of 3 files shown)


Frodigo (approved):
Hi @marc-arnoult, good catch! Thanks for that.

PR #760, opened 2022-03-21
refactor: refactored the useApi composable

Frodigo:
## Description
The first step of the useApi composable refactoring:

## Motivation and Context
providing better developer experience and open opportunity to use the useApi composable to query data in more generic way.

## How Has This Been Tested?
tested on my local env

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/composables/useApi/index.ts
+++ b/packages/theme/composables/useApi/index.ts
@@ -1,6 +1,23 @@
-import { useContext } from '@nuxtjs/composition-api';
+import { useContext, ref } from '@nuxtjs/composition-api';
 import { request as graphQLRequest, GraphQLClient } from 'graphql-request';
+import { Logger } from '@vue-storefront/core';
 import cookieNames from '~/enums/cookieNameEnum';
+
+export type RequestParams<VARIABLES> = {
+  document: string,
+  variables?: VARIABLES | null,
+  endpoint?: string,
+  requestHeaders?: {
+    authorization?: string,
+    store?: string
+  },
+  debugInfo?: string
+};
+
+enum GraphQLOperation {
+  query = 'query',
+  mutation = 'mutation',
+}
 
 export const useApi = () => {
   const { app } = useContext();
@@ -14,6 +31,8 @@ export const useApi = () => {
     authorization?: string,
     store?: string
   } = {};
+  const error = ref(null);
+  const loading = ref(false);
 
   if (customerToken) {
     defaultHeaders.authorization = `Bearer ${customerToken}`;
@@ -23,26 +42,60 @@ export const useApi = () => {
     defaultHeaders.store = storeCode;
   }
 
-  const query = (
-    document: string,
-    variables = null,
-    endpoint = defaultEndpoint,
-    requestHeaders = defaultHeaders,
-  ) => new GraphQLClient(endpoint, {
-    method: useGETForQueries ? 'GET' : 'POST',
-    headers: requestHeaders,
-  }).request(document, variables);
+  const request = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>, type: string): Promise<TYPE> => {
+    const params: RequestParams<VARIABLES> = {
+      variables: null,
+      endpoint: defaultEndpoint,
+      requestHeaders: defaultHeaders,
+      debugInfo: '',
+      ...requestParams,
+    };
+    let response = null;
 
-  const mutate = (
-    document: string,
-    variables = null,
-    endpoint = defaultEndpoint,
-    requestHeaders = defaultHeaders,
-  ) => graphQLRequest(endpoint, document, variables, requestHeaders);
+    try {
+      Logger.debug(`[Magento] ${params.debugInfo}`, { params });
+      loading.value = true;
+
+      switch (type) {
+        case GraphQLOperation.query: {
+          response = new GraphQLClient(params.endpoint, {
+            method: useGETForQueries ? 'GET' : 'POST',
+            headers: params.requestHeaders,
+          }).request<TYPE, VARIABLES>(params.document, params.variables);
+
+          break;
+        }
+
+        case GraphQLOperation.mutation: {
+          response = graphQLRequest<TYPE, VARIABLES>(params.endpoint, params.document, params.variables, params.requestHeaders);
+
+          break;
+        }
+        default: {
+          break;
+        }
+      }
+
+      Logger.debug(`[Magento] ${params.debugInfo} – results: `, { response });
+    } catch (err) {
+      error.value = err;
+      Logger.error(`[Magento] ${params.debugInfo} – error: `, { response });
+    } finally {
+      loading.value = false;
+    }
+
+    return response;
+  };
+
+  const query = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>): Promise<TYPE> => request(requestParams, GraphQLOperation.query);
+
+  const mutate = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>): Promise<TYPE> => request(requestParams, GraphQLOperation.mutation);
 
   return {
     query,
     mutate,
+    error,
+    loading,
   };
 };
 
--- a/packages/theme/components/TopBar/TopBar.d.ts
+++ b/packages/theme/components/TopBar/TopBar.d.ts
@@ -1,0 +1,6 @@
+import { AvailableStores, Currency } from '~/composables/types';
+
+export type CheckStoresAndCurrencyQueryResponse = {
+  availableStores: AvailableStores,
+  currency: Currency
+};
--- a/packages/theme/components/TopBar/useTopBar.ts
+++ b/packages/theme/components/TopBar/useTopBar.ts
@@ -2,6 +2,7 @@ import { ref, onMounted } from '@nuxtjs/composition-api';
 
 import useApi from '~/composables/useApi';
 import checkStoresAndCurrencyQuery from './checkStoresAndCurrency.gql';
+import type { CheckStoresAndCurrencyQueryResponse } from './TopBar';
 
 export const useTopBar = () => {
   const { query } = useApi();
@@ -9,7 +10,10 @@ export const useTopBar = () => {
   const hasCurrencyToSelect = ref(null);
 
   onMounted(() => {
-    query(checkStoresAndCurrencyQuery)
+    query<CheckStoresAndCurrencyQueryResponse, null>({
+      document: checkStoresAndCurrencyQuery,
+      debugInfo: 'useTopBar - Check stores and currency',
+    })
       .then((data) => {
         hasStoresToSelect.value = data?.availableStores?.length;
         hasCurrencyToSelect.value = data?.currency?.available_currency_codes?.length > 1;
--- a/packages/theme/modules/catalog/category/components/cms/useCategoryContent.ts
+++ b/packages/theme/modules/catalog/category/components/cms/useCategoryContent.ts
@@ -1,18 +1,15 @@
 import useApi from '~/composables/useApi';
 import categoryContentQuery from './categoryContent.gql';
 import displayModesEnum from '~/modules/catalog/category/enums/displayModesEnum';
+import type { CategoryTree } from '~/composables/types';
+import type { CmsBlock } from '~/modules/GraphQL/types';
 
-export interface CmsBlockInterface {
-  /** CMS block content */
-  content?: string;
-  /** CMS block identifier */
-  identifier?: string;
-  /** CMS block title */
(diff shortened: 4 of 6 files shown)


sethidden (changes requested):
I think this approach creates an additional level of abstraction (in the form of the `request` without offering much benefit in return. query and mutate are `request`s on their own, I don't think there's a reason to wrap them like that.

What are the problems with the current implementation?

I understand the request() and the switch is made to not to double the error.value and loading.value assignments but as a side effect, but here I'd really just prefer the duplication (not all duplication is bad) like we have in all the other composables we've ported (error.value repeating a lot etc)

---

I'm not super keen on keeping the useApi composable tbh, I'd much prefer to leave it as is for now. This is a thing that seems like it'd be best to discuss in a meeting, add it to the backlog, measure it on the refinement etc.

packages/theme/composables/useApi/index.ts
```diff
@@ -1,7 +1,24 @@
-import { useContext } from '@nuxtjs/composition-api';
+import { useContext, ref } from '@nuxtjs/composition-api';
 import { request as graphQLRequest, GraphQLClient } from 'graphql-request';
+import { Logger } from '@vue-storefront/core';
 import cookieNames from '~/enums/cookieNameEnum';
 
+export type RequestParams<VARIABLES> = {
+  document: string,
+  variables?: VARIABLES | null,
```
bartoszherba, line 8: is the "or NULL" REALLY important to have? If not passed it will be falsy anyway

packages/theme/composables/useApi/index.ts
```diff
@@ -23,26 +42,60 @@ export const useApi = () => {
...
-    headers: requestHeaders,
-  }).request(document, variables);
-
-  const mutate = (
-    document: string,
-    variables = null,
-    endpoint = defaultEndpoint,
-    requestHeaders = defaultHeaders,
-  ) => graphQLRequest(endpoint, document, variables, requestHeaders);
+  const request = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>, type: string): Promise<TYPE> => {
```
bartoszherba, line 45:
```suggestion
  const request = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>, type: string = GraphQLOperation.query): Promise<TYPE> => {
```

packages/theme/composables/useApi/index.ts
```diff
@@ -23,26 +42,60 @@ export const useApi = () => {
...
+        }
+
+        case GraphQLOperation.mutation: {
+          response = graphQLRequest<TYPE, VARIABLES>(params.endpoint, params.document, params.variables, params.requestHeaders);
+
+          break;
+        }
+        default: {
+          break;
+        }
```
bartoszherba, line 76:
```suggestion
```

Frodigo:
> I think this approach creates an additional level of abstraction (in the form of the `request` without offering much benefit in return. query and mutate are `request`s on their own, I don't think there's a reason to wrap them like that.
> 
> What are the problems with the current implementation?

Current implementation of useApi itself is ok, but that changes resolves other problem. 

Now, we have ~ 35 composables in theme/composables directory, and a lot of them are redundant because the only change that is dynamic there is a graphql query. 

Typically there are three fields:
- error
- loading
- load or search

So we are repeating code as monkeys, and the difference is only the name of composable (for example useRelatedProducts, useUpsellProducts, useCategory, useCategoryContent and so on)

We can delegate that logic to useApi, and remove redundant composables. I wanted to do that in the next step. 

On the other hand, we need to maintain that redundant composables code and if we had wanted to cover composables by unit tests we would have to repeat ourselves again in tests. In the new approach we have less code, fewer tests, and fewer places to maintain.

> 
> I understand the request() and the switch is made to not to double the error.value and loading.value assignments but as a side effect, but here I'd really just prefer the duplication (not all duplication is bad) like we have in all the other composables we've ported (error.value repeating a lot etc)

Composables you mentioned are not using this useApi composable. Those composables use API client from `@vue-storefront/magento-api. 

In next step we would refactor composables to new API, and avoid duplications.

> 
> I'm not super keen on keeping the useApi composable tbh, I'd much prefer to leave it as is for now. This is a thing that seems like it'd be best to discuss in a meeting, add it to the backlog, measure it on the refinement etc then do it in an organized way like we've done with the composables
> 
> I'm thinking this refactor could be better if request() took a function that returns a promise as first argument instead of the switch statement

I don't understand. The request function purpose is to send a query or mutation and prepare correct request parameters. I created the request function only because I didn't want to repeat code in both query and mutate functions. 

The switch statement is only for deciding which graphql operation should be sent: query or mutation, moreover the request function is private.

packages/theme/composables/useApi/index.ts
```diff
@@ -23,26 +42,60 @@ export const useApi = () => {
...
-
-  const mutate = (
-    document: string,
-    variables = null,
-    endpoint = defaultEndpoint,
-    requestHeaders = defaultHeaders,
-  ) => graphQLRequest(endpoint, document, variables, requestHeaders);
+  const request = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>, type: string): Promise<TYPE> => {
+    const params: RequestParams<VARIABLES> = {
+      variables: null,
```
bartoszherba, line 47: Field is not required
```suggestion
```

packages/theme/composables/useApi/index.ts
```diff
@@ -23,26 +42,60 @@ export const useApi = () => {
...
+      endpoint: defaultEndpoint,
+      requestHeaders: defaultHeaders,
+      debugInfo: '',
```
bartoszherba, line 50: Field is not required
```suggestion
```

sethidden:
@Frodigo Down the line, I think using useApi for the error.value and loading.value logic is ok as it reduces the code duplication you mentioned, but adding `request` which wraps `query` and `mutation` shouldn't be included here, as `query` and `mutation` and cookies etc. shouldn't be in `useApi` in the first place (need to move cookies out of context etc). So to me it's necessary to take 2 steps back to take the step you want to take here

Let's see what comes out of our call soon :P

Frodigo:
> @Frodigo Down the line, I think using useApi for the error.value and loading.value logic is ok as it reduces the code duplication you mentioned, but adding `request` which wraps `query` and `mutation` shouldn't be included here, as `query` and `mutation` and cookies etc. shouldn't be in `useApi` in the first place (need to move cookies out of context etc). So to me it's necessary to take 2 steps back to take the step you want to take here
> 
> Let's see what comes out of our call soon :P

So you would remove cookies state from the useApi composable? 

As I mentioned before the `request` method is private and it's only for remove code duplication in query and mutate functions. I didn't want to repeat things like params creation and try catch block, logging etc.

Without request function code would look like this:


```
const mutate = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>, type: string): Promise<TYPE> => {
    const params: RequestParams<VARIABLES> = {
      variables: null,
      endpoint: defaultEndpoint,
      requestHeaders: defaultHeaders,
      debugInfo: '',
      ...requestParams,
    };
    let response = null;

    try {
      Logger.debug(`[Magento] ${params.debugInfo}`, { params });
      loading.value = true;

      response =  graphQLRequest<TYPE, VARIABLES>(params.endpoint, params.document, params.variables, params.requestHeaders);

      Logger.debug(`[Magento] ${params.debugInfo} – results: `, { response });
    } catch (err) {
      error.value = err;
      Logger.error(`[Magento] ${params.debugInfo} – error: `, { response });
    } finally {
      loading.value = false;
    }

    return response;
  };


const query = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>, type: string): Promise<TYPE> => {
    const params: RequestParams<VARIABLES> = {
      variables: null,
      endpoint: defaultEndpoint,
      requestHeaders: defaultHeaders,
      debugInfo: '',
      ...requestParams,
    };
    let response = null;

    try {
      Logger.debug(`[Magento] ${params.debugInfo}`, { params });
      loading.value = true;

      response = new GraphQLClient(params.endpoint, {
            method: useGETForQueries ? 'GET' : 'POST',
            headers: params.requestHeaders,
          }).request<TYPE, VARIABLES>(params.document, params.variables);

      Logger.debug(`[Magento] ${params.debugInfo} – results: `, { response });
    } catch (err) {
      error.value = err;
      Logger.error(`[Magento] ${params.debugInfo} – error: `, { response });
    } finally {
      loading.value = false;
    }

    return response;
  };
```

packages/theme/composables/useApi/index.ts
```diff
@@ -1,7 +1,24 @@
-import { useContext } from '@nuxtjs/composition-api';
+import { useContext, ref } from '@nuxtjs/composition-api';
```
bartoszherba, line 1: In general. I think that API client do not have to utilize composition api so it will not share more technical debt than necessary. Eg. loading or errors as ref are not really required. Loading can be easily handled by the client and error can be simply thrown so the client can handle it. This is not a big thing maybe, just the less debt risk the better.

bartoszherba (changes requested):

packages/theme/composables/useApi/index.ts
Frodigo: Good point,

so let me understand correctly that:
1. Instead of a return error that is a Ref, return a plain error? 
2. loading field is not necessary here, we remove it, and components, when useApi is used, takes care of handle loading state?

Frodigo: @bartoszherba  In addition to point 1. Maybe we can just throw an error instead of returning an error field? cc: @sethidden

sethidden (changes requested):
Let's talk about this on the daily meeting

packages/theme/composables/useApi/index.ts
sethidden: If error and loading weren't refs then using useApi would be no different than axios.get()

e.g. loading is a good example - why litter the template with const isXLoading = ref(false), const isYLoading = ref(false) if that can be returned from an useApi? Perhaps Bartek is talking about a different form of useApi (I'm talking about the "simple promise helper" form)

sethidden:
image
Would it make sense to just use config.get instead of this? This would further detach useApi from the context.

At this point $magento.config is a redundant extra layer of abstraction over just what we have in config.get. I guess it's ok for useApi to still use config.get or CookiesUniversal.get (instead of $magento.config.state.get[CookieName]) even if this is inconsistent - independence from context is king

Frodigo:
> image
> 
> Would it make sense to just use config.get instead of this? This would further detach useApi from the context.
> At this point $magento.config is a redundant extra layer of abstraction over just what we have in config.get. I guess it's ok for useApi to still use config.get or CookiesUniversal.get (instead of $magento.config.state.get[CookieName]) even if this is inconsistent - independence from context is king

I tried to use cookie CookiesUniversal.get but app crashed then. 

Here is the code snippet: 

```
import Cookie from 'cookie-universal';
import { request as graphQLRequest, GraphQLClient } from 'graphql-request';
import cookieNames from '~/enums/cookieNameEnum';

export type RequestParams<VARIABLES> = {
  document: string,
  variables?: VARIABLES | null,
  endpoint?: string,
  requestHeaders?: {
    authorization?: string,
    store?: string
  },
  debugInfo?: string
};

const defaultEndpoint = process.env.MAGENTO_GRAPHQL;

const buildParams = <VARIABLES>(requestParams: RequestParams<VARIABLES>): RequestParams<VARIABLES> => {
  const defaultHeaders: {
    authorization?: string,
    store?: string
  } = {};

  const cookies = Cookie();
  const customerToken = cookies.get(cookieNames.customerCookieName);
  const storeCode = cookies.get(cookieNames.storeCookieName);

  if (customerToken) {
    defaultHeaders.authorization = `Bearer ${customerToken}`;
  }

  if (storeCode) {
    defaultHeaders.store = storeCode;
  }

  return {
    endpoint: defaultEndpoint,
    requestHeaders: {
      ...defaultHeaders,
      ...requestParams.requestHeaders,
    },
    ...requestParams,
  };
};

export const mutate = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>): Promise<TYPE> => {
  const params = buildParams<VARIABLES>(requestParams);

  return graphQLRequest<TYPE, VARIABLES>(params.endpoint, params.document, params.variables, params.requestHeaders);
};

export const query = <TYPE, VARIABLES>(requestParams: RequestParams<VARIABLES>): Promise<TYPE> => {
  const params = buildParams<VARIABLES>(requestParams);
  console.log(requestParams);

  return new GraphQLClient(params.endpoint, {
    method: 'GET',
    headers: params.requestHeaders,
  }).request<TYPE, VARIABLES>(params.document, params.variables);
};
```

PR #771, opened 2022-03-22
chore: add more GitHub environments for more detailed Jira deployments

ghost:
I tested this by including this detailed-environments branch to the workflow .yml files in the on: push: branches: properties (they're now removed because I obviously don't want to commit them to develop). 

The succesful workflow run is here: https://github.com/vuestorefront/magento2/actions/runs/2027541890

You can tell  it works because the "enterprise" environment appeared in github
image

This is so Marcin doesn't have to copy-paste the GitHub action deployment link in every task that gets merged - the deployments for dev,canary etc. envs should be available from within the JIRA deployments tab

image

Right now we just have a single "production" environment, but really underneath it's 4 different environments under the "production" umbrella.

The goal of this task is to create 4 individual, independent environments within GitHub and use the deployment API to show the status of each environment separately, not under the production umbrella.

For example, see https://vsf.atlassian.net/jira/software/c/projects/M2/deployments?startDate=-3m&endDate=now and all of the deployments are "production", even though we haven't deployed to the production env at demo-magento2 in quite a while.

See https://github.com/atlassian/github-for-jira/issues/772

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/deploy-vue-storefront-cloud.yml
+++ b/.github/workflows/deploy-vue-storefront-cloud.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - develop
+      - enterprise
       - release/*
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/deployment-template.yml
```diff
@@ -0,0 +1,67 @@
...
+      - uses: chrnorm/deployment-action@releases/v1
+        name: Create GitHub deployment
+        id: deployment
+        with:
+          token: ${{ github.token }}
+          environment: ${{ inputs.github_environment_name }}
+          initial_status: in_progress
+          target_url: ${{ inputs.target_url }}
+      - name: Deploy on ${{ inputs.target_url }}
+        run: |
```
sethidden, line 36: I've moved the previously duplicated scripts to a reusable workflow file

.github/workflows/deploy-vue-storefront-cloud.yml
```diff
@@ -47,86 +48,52 @@ jobs:
...
+    uses: ./.github/workflows/deployment-template.yml
     needs: build
-    steps:
-      - uses: chrnorm/deployment-action@releases/v1
-        name: Create GitHub deployment
-        id: deployment
-        with:
-          token: ${{ github.token }}
-          target_url: https://demo-magento2.europe-west1.gcp.storefrontcloud.io
-          environment: production
```
sethidden, line 60: This is the culprit of why all deployments are understood as "production" instead of being more granular (dev, canary etc. envs that we have) - it's just always pushing to the GitHub deployment api saying it's deploying production

.github/workflows/deploy-vue-storefront-cloud.yml
```diff
@@ -47,86 +48,52 @@ jobs:
           SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
           RECAPTCHA_SITE_KEY: ${{ secrets.RECAPTCHA_SITE_KEY }}
           RECAPTCHA_SECRET_KEY: ${{ secrets.RECAPTCHA_SECRET_KEY }}
-  deploy:
-    runs-on: ubuntu-latest
+
+  deploy-main:
```
sethidden, line 52: Before someone reviews saying "why are there so many deployment jobs - there should be just one job that has many `steps` that are conditional"

unfortunately github does not support reusable workflows (`uses: ./.github/workflows/deployment-template`) in steps - they always have to be jobs

.github/workflows/deploy-vue-storefront-cloud.yml
```diff
@@ -47,86 +48,52 @@ jobs:
...
+    secrets:
+      cloud_username: ${{ secrets.CLOUD_USERNAME }}
+      cloud_password: ${{ secrets.CLOUD_PASSWORD }}
+      
+  deploy-develop:
+    uses: ./.github/workflows/deployment-template.yml
+    needs: build
+    if: github.ref == 'refs/heads/develop'
+    with: 
+      github_environment_name: dev
```
sethidden, line 69: Just the github environment name, I 've already set them up - they just don't appear on GitHub because they've never been pushed to (except for "production" and enterprise which I used for testing)

image

RetroModernDev (approved):

Issue #766, opened 2022-03-22
[Bug]: Customer Address - update customer address region issue


bartoszherba:
### Expected Behaviour
* I am able to update/create customer address including all available fields in particular
* when I try to change region details I am able to update the address
* when i select country with available predefined regions the select is properly populated and visible

### Actual Behaviour
* When I try to change region details I am not able to update the address.
* When I select a country with predefined regions I do not see the list nor the select with regions

### Steps to reproduce
1. Log in as a customer
1. Go to the customer account -> address details
1. Try to change any address (immediately observe that address is loaded with Argentina region - or any first available)
1. Change country and one irrelevant field for the test
1. Try to save new address
1. Observe errors in the console, address was not updated (ref. to the updated field)
1. Also, eg. for Poland there is no select for regions

### What version of Magento 2 integration are you using?
2.4.x

### What version of Node.js are you using?
16.13.1

### What browser (and version) are you using?
Version 96.0.4664.110 (Official Build) (arm64)

### What operating system (and version) are you using?
macOS Monterey 12.1

### Relevant log output
```shell
[VSF][error]:  [GraphQL error]: Message: A region_id is required for the specified country code, Location: [column: 3, line: 2], Path: updateCustomerAddress
```

### Code of Conduct
* [x] I agree to follow this project's Code of Conduct

Issue #765, opened 2022-03-22
[Bug]: CustomerAccount - deletion of address is not reflected in the UI


bartoszherba:
### Expected Behavior
When I delete the address in my Address Book it should be removed from the list.

### Actual Behavior
When I delete the address in my Address Book it is not removed from the list.

### Steps to reproduce
1. Log in as a customer
1. Go to you address book
1. Assuming you have addresses that can be deleted, delete one
1. Observe that the address is not removed from the list
1. Try to delete it again
1. In the console there will be an error that the address with ID "x" is not found
1. Reload page and observe that the address was actually deleted

### What version of Magento 2 integration are you using?
2.4.x

### What version of Node.js are you using?
16.13.1

### What browser (and version) are you using?
Version 96.0.4664.110 (Official Build) (arm64)

### What operating system (and version) are you using?
macOS Monterey 12.1

### Code of Conduct
* [x] I agree to follow this project's Code of Conduct

PR #776, opened 2022-03-23
fix: fixed adding/removing address funcionality

Frodigo:
## Description
- fixed adding address on my account
- fixed removing address on my account

## Related Issue
#765 #766

## Motivation and Context
bugfixing

## How Has This Been Tested?
tested on my account:
1. log-in as a customer
2. go to address details
3. add a new address
4. delete address

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/components/MyAccount/AddressForm.vue
+++ b/packages/theme/components/MyAccount/AddressForm.vue
@@ -140,7 +140,7 @@
             required
             :valid="!errors[0]"
             :error-message="$t(errors[0])"
-            @input="searchCountry({ id:$event })"
+            @input="reloadCountry({ id: $event })"
           >
             <SfSelectOption
               v-for="countryOption in countriesList"
@@ -273,6 +273,10 @@ export default defineComponent({
     const countries = ref([]);
     const country = ref(null);
 
+    const reloadCountry = async (params) => {
+      country.value = await searchCountry(params);
+    };
+
     const form = reactive({
       apartment: props.address.apartment,
       city: props.address.city,
@@ -320,7 +324,7 @@ export default defineComponent({
       submitForm,
       countriesList,
       regionInformation,
-      searchCountry,
+      reloadCountry,
     };
   },
 });
--- a/packages/theme/pages/MyAccount/AddressesDetails.vue
+++ b/packages/theme/pages/MyAccount/AddressesDetails.vue
@@ -120,6 +120,16 @@ export default defineComponent({
     const getTranslatedUrlAddress = (title) => app.i18n.t(`${title}`).toLowerCase().replace(' ', '-');
     const isNewAddress = computed(() => route.value.query.id === 'new');
     const editingAddress = computed(() => !!route.value.query.id);
+    const { fetch } = useFetch(async () => {
+      const addressesData = await load();
+      userAddresses.value = userAddressesGetters.getAddresses(addressesData);
+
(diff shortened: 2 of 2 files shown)


packages/theme/components/MyAccount/AddressForm.vue
```diff
@@ -273,6 +273,10 @@ export default defineComponent({
     const countries = ref([]);
     const country = ref(null);
 
+    const reloadCountry = async (params) => {
```
bartoszherba, line 276: I am not sure if this should block further script execution.

bartoszherba (approved):
One comment, nothing critical I think.

packages/theme/components/MyAccount/AddressForm.vue
sethidden: This is a function that awaits the promise only if it's called, it'll run only on user input after the initial hydration finished

packages/theme/pages/MyAccount/AddressesDetails.vue
```diff
@@ -120,6 +120,16 @@ export default defineComponent({
     const getTranslatedUrlAddress = (title) => app.i18n.t(`${title}`).toLowerCase().replace(' ', '-');
     const isNewAddress = computed(() => route.value.query.id === 'new');
     const editingAddress = computed(() => !!route.value.query.id);
+    const { fetch } = useFetch(async () => {
+      const addressesData = await load();
+      userAddresses.value = userAddressesGetters.getAddresses(addressesData);
+
+      const activeAddressData = userAddresses.value
+        .filter((address) => String(address?.id) === route.value.query.id)
+        .pop();
```
sethidden, line 129:
```suggestion
      const [ ..._front, activeAddressData] = userAddresses.value
        .filter((address) => String(address?.id) === route.value.query.id);
```

```suggestion
      const [ ...front, activeAddressData] = userAddresses.value
        .filter((address) => String(address?.id) === route.value.query.id).at(-1);
```

.pop() seems ununtuitive to me here, proposed some more explicit ways

packages/theme/pages/MyAccount/AddressesDetails.vue
```diff
@@ -135,7 +145,7 @@ export default defineComponent({
       const isDefault = userAddressesGetters.isDefault(address);
       if (!isDefault) {
         await remove({ address });
-        await load({});
+        fetch();
```
sethidden, line 148:
```suggestion
        await fetch();
```

Is dropping the await intentional here? Could be problematic if someone added something new behind the fetch later

sethidden (changes requested):

packages/theme/pages/MyAccount/AddressesDetails.vue
Frodigo: thanks

packages/theme/pages/MyAccount/AddressesDetails.vue
Frodigo: fetch retriggers useFetch and it does not return a promise so it cannot be awaited.

packages/theme/pages/MyAccount/AddressesDetails.vue
Frodigo: but after change you requested code stopped work as excepted

sethidden (approved):

PR #780, opened 2022-03-23
fix: remove commitizen prepare-commit-msg git hook

ghost:
# STOP - Actually check out to the branch, run "yarn" and see if commiting to the branch the way you usually do it (maybe from within IDE? maybe terminal?) behaves to your liking

We have a git hook that launches when you type `git commit`, the problem up until now is that it just errored out during launch, so it didn't change the way git commit works originally. Here I'm fixing it so it works as intended by the original author of the commit that added it.

But I suspect that the sudden change of how `git commit` works may break the workflows of contributors and devs who commit outside the terminal. Basically, please check out to this branch and try to commit something the way you usually do it and see if that isn't annoying. You can commit from inside Webstorm, using the terminal, or from Vim (like I do and where this PR breaks things, so I'll have to commit from the terminal if this is merged)

## I'm super open to removing this hook altogether

### DIFF
--- a/.husky/prepare-commit-msg
+++ b/.husky/prepare-commit-msg
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-exec < /dev/tty && git cz --hook || true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ To make your life easier @vuestorefront/magento2 is commitizen-friendly and prov
 > Commit standards
 
 - [conventional-changelog](https://github.com/conventional-changelog)
-- husky commit message hook available
+- run `npx cz` to commit using commitizen
 - present tense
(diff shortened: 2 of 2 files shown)


sethidden:
https://user-images.githubusercontent.com/5359825/159714779-627de109-2c37-4db2-85c6-b82e9ade901f.mov

  a

bloodf (dismissed):

roziscoding:
I'm also in favor of removing this hook, or maybe finding a way of making it optional. If people don't run `husky install` then the hook shouldn't be created, right?

Anyway, commitizen doesn't actually guarantee good commit messages, commitlint does that, so the hook is more of a "cosmetic" feature.

Frodigo:
I am either voting for removing that pre-commit hook and as you mentioned @sethidden - keeping `cz` in package.json is a good idea.

RetroModernDev (approved):

PR #786, opened 2022-03-24
refactor: use in-house logger over @vue-storefront-core's

ghost:
This is basically a 70% rewrite of the @vue-storefront/core logger, just copied over to the magento2 repo (as part of our process of migrating away from core). I didn't want to copy the core one as it had a bit too much branching out functions.

The problem I've noticed is that every other integration team that's migrating away from core (and disabling it's features one-by-one -> logger, context, etc) will end up having duplicated logger code  (so there'll be ~5 logger implementations across vue-storefront) - each team will have to:
1. disable the `@vue-storefront/core` logger plugin in nuxt.config.js ({logger: false})
2. write their own logger (probably inspierd by core's code, like I did here)
3. replace the imports from core to their own logger

I'm wondering if it'd make sense to publish a modified (more extendable) version of this as a separate package? e.g. `@vue-storefront/logger`. `@vue-storefront/core` is very "use the whole thing or don't use it at all", whereas `@vue-storefront/logger` would just be this tiny snippet that does one thing only.

### DIFF
--- a/packages/theme/composables/useAddresses/index.ts
+++ b/packages/theme/composables/useAddresses/index.ts
@@ -1,5 +1,5 @@
 import { Ref, ref, useContext } from '@nuxtjs/composition-api';
-import { Logger } from '@vue-storefront/core';
+import { Logger } from '~/helpers/logger';
 import { transformUserCreateAddressInput, transformUserUpdateAddressInput } from '~/helpers/userAddressManipulator';
 import { ComposableFunctionArgs } from '~/composables/types';
 import { UseAddressesInterface, UseAddressesParamsInput, UseAddressesErrors } from '~/composables/useAddresses/useAddresses';
--- a/packages/theme/composables/useBilling/commands/saveBillingAddressCommand.ts
+++ b/packages/theme/composables/useBilling/commands/saveBillingAddressCommand.ts
@@ -1,4 +1,4 @@
-import { Logger } from '@vue-storefront/core';
+import { Logger } from '~/helpers/logger';
 import { BillingCartAddress, Maybe, SetBillingAddressOnCartInput } from '~/modules/GraphQL/types';
 
 export const saveBillingAddressCommand = {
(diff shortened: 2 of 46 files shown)


packages/theme/plugins/logger/index.ts
```diff
@@ -0,0 +1,8 @@
+import { getLoggerWithVerbosity, getVerbosity } from './verbosity';
+
+const verbosity = getVerbosity(process.env.NODE_ENV);
+export const Logger = getLoggerWithVerbosity(verbosity);
+
+export default (_, inject) => {
+  inject('logger', Logger);
```
sethidden, line 7: @Frodigo You've mentioned in the Jira issue that you'd like the logger to be registered as a plugin. But since we'll be moving away from keeping axios in the context (moving it to a normal import) as well, would it be okay if I just didn't register this plugin here?

ie. everytime you wanted to log something you just import the Logger. I've replaced the path in 40 files and never saw any file using the logger in the context. I don't think @vue-storefront/core injects the logger as a plugin either.

Would that be ok for you?

bloodf:
Creating a new package on the core makes more sense than adding to the integration. It's a separation of responsibilities. We have the integration doing what it knows best and creating a new package on the core that will handle the login because the login can be extended in the future to accept other stuff or do something more complex. 

The DX team could prepare something if you have a proposition on what you expect to achieve in this separation.

packages/theme/plugins/logger/verbosity.ts
```diff
@@ -0,0 +1,43 @@
+import { getStyledConsolePrefix } from './style';
+
+export type LogEntryVariant = 'error' | 'warn' | 'info' | 'debug' | 'none';
+
+const partial = (fn: Function, ...args: unknown[]) => (...laterArgs: unknown[]) => { console.debug(args, laterArgs); fn(...args, ...laterArgs); };
```
sethidden, line 5:
```suggestion
const partial = (fn: Function, ...args: unknown[]) => (...laterArgs: unknown[]) => fn(...args, ...laterArgs);
```

packages/theme/plugins/logger/index.ts
Frodigo: We don't use Logger as a plugin so it's ok.

RetroModernDev (dismissed):

RetroModernDev (approved):

PR #790, opened 2022-03-25
refactor: replaced onSSR method calls wih Nuxt useAsync and useFetch hooks

Frodigo:
## Description
- removed onSSR hooks from our templates
- fixed broken my account page

## Motivation and Context
refactorization, bugfixing

## How Has This Been Tested?
tested on my local, for logged-in and not logged customer

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/components/AppHeader.vue
+++ b/packages/theme/components/AppHeader.vue
@@ -161,20 +161,21 @@ export default defineComponent({
     const { setTermForUrl, getAgnosticCatLink } = useUiHelpers();
     const { isAuthenticated } = useUser();
     const { loadTotalQty: loadCartTotalQty, cart } = useCart();
-    const { loadItemsCount: loadWishlistItemsCount, wishlist } = useWishlist();
+    const { loadItemsCount: loadWishlistItemsCount } = useWishlist();
 
     const { categories: categoryList, search: categoriesListSearch } = useCategory('AppHeader:CategoryList');
 
     const isSearchOpen = ref(false);
     const result = ref(null);
+    const wishlistItemsQty = ref(null);
 
-    const wishlistHasProducts = computed(() => wishlist.value.items_count > 0);
+    const wishlistHasProducts = computed(() => wishlistItemsQty.value > 0);
     const accountIcon = computed(() => (isAuthenticated.value ? 'profile_fill' : 'profile'));
     const categoryTree = ref([]);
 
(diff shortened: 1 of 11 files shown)


packages/theme/components/AppHeader.vue
```diff
@@ -190,7 +191,13 @@ export default defineComponent({
     onMounted(() => {
       if (app.$device.isDesktop) {
         loadCartTotalQty();
-        loadWishlistItemsCount();
+        // eslint-disable-next-line promise/catch-or-return
```
sethidden, line 194: Possible to use async/await here instead of old-school then()?
```
async onMounted() {
  wishlistItemQty.value = await loadWishlistItemsCount()
}
```

packages/theme/components/AppHeader.vue
```diff
@@ -190,7 +191,13 @@ export default defineComponent({
...
+        loadWishlistItemsCount()
+          .then((response) => {
+            wishlistItemsQty.value = response;
+
+            return response;
```
sethidden, line 199: If not, there's no point in returning anything here. Just the assignment is necessary

packages/theme/components/WishlistSidebar.vue
```diff
@@ -220,8 +223,14 @@ export default defineComponent({
     const { getMagentoImage, imageSizes } = useImage();
 
     onMounted(() => {
-      if (!wishlist.value.id) {
-        loadWishlist();
+      if (!customerStore.wishlist.id) {
+        // eslint-disable-next-line promise/catch-or-return
```
sethidden, line 227: Same here (both the extraneous return and async await)

sethidden (changes requested):

packages/theme/components/AppHeader.vue
Frodigo: nope, because I don't want to block rendering.

packages/theme/components/AppHeader.vue
Frodigo: Our eslint rules say that there must be the value returned so if we want to change that, first verify and change linting rules.

packages/theme/components/WishlistSidebar.vue
Frodigo: As above

sethidden (approved):

sethidden (approved):

PR #791, opened 2022-03-25
chore: remove context shim

ghost:
this was necessary when we still had the @vue-storefront/core
composables in packages/theme

disabling the @vue-storefront/core context plugin was not possible
without the shim from this commit

since we've already migrated all composables this is ok to remove


packages/theme/pages/Checkout/Payment.vue
```diff
@@ -126,7 +126,7 @@
           <template #label>
             <div class="sf-checkbox__label">
               {{ $t('I agree to') }}
-              <SfLink href="#">
```
sethidden, line 129: just a "by-the-way" fix - link is a required prop and throws an error in console if it's not set

RetroModernDev (approved):

PR #797, opened 2022-03-30
refactor: unify cookie retrieval

sethidden:
Up until now we've had two cookie APIs: $cookies.get/set/remove and the vue-storefront/core provided $vsf.config.state.set[CookieName]. This PR removes all instances of the first API and makes them use the $vsf.config.state API

### DIFF
--- a/packages/api-client/src/helpers/apiClient/defaultSettings.ts
+++ b/packages/api-client/src/helpers/apiClient/defaultSettings.ts
@@ -23,6 +23,9 @@ export const defaultSettings: ClientConfig = {
     setLocale: () => {},
     getCountry: () => '',
     setCountry: () => {},
+    setMessage: () => {},
+    // @ts-ignore
+    getMessage: () => ({}),
   },
   externalCheckout: {
     enable: false,
--- a/packages/theme/composables/useMagentoConfiguration.ts
+++ b/packages/theme/composables/useMagentoConfiguration.ts
@@ -12,20 +12,19 @@ type UseMagentoConfiguration = () => {
   loadConfiguration: (params: { updateCookies: boolean; updateLocale: boolean; }) => Promise<void>;
 };
 
-// @ts-ignore
 export const useMagentoConfiguration: UseMagentoConfiguration = () => {
-  const { app } = useContext();
+  const { app: { i18n, $vsf: { $magento: { config } } } } = useContext();
 
   const {
     config: storeConfig,
     load: loadConfig,
   } = useConfig();
 
-  const selectedCurrency = computed<string | undefined>(() => app.$vsf.$magento.config.state.getCurrency());
-  const selectedLocale = computed<string | undefined>(() => app.$vsf.$magento.config.state.getLocale());
-  const selectedStore = computed<string | undefined>(() => app.$vsf.$magento.config.state.getStore());
+  const selectedCurrency = computed<string | undefined>(() => config.state.getCurrency());
+  const selectedLocale = computed<string | undefined>(() => config.state.getLocale());
+  const selectedStore = computed<string | undefined>(() => config.state.getStore());
 
-  const loadConfiguration: (params: { updateCookies: boolean; updateLocale: boolean; }) => void = (params = {
+  const loadConfiguration: (params: { updateCookies: boolean; updateLocale: boolean; }) => Promise<void> = async (params = {
     updateCookies: false,
     updateLocale: false,
   }) => {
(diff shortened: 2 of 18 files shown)


packages/api-client/src/helpers/apiClient/defaultSettings.ts
```diff
@@ -23,6 +23,9 @@ export const defaultSettings: ClientConfig = {
     setLocale: () => {},
     getCountry: () => '',
     setCountry: () => {},
+    setMessage: () => {},
+    // @ts-ignore
```
sethidden, line 27: I can't get TS to be quiet about this. This is not an important file as it's just the defaultSettings and a function that returns void

packages/theme/composables/useUiNotification/index.ts
```diff
@@ -34,7 +34,7 @@ const useUiNotification = () => {
 
       if (index !== -1) state.notifications.splice(index, 1);
 
-      app.$cookies.remove(cookieNames.messageCookieName);
+      app.$vsf.$magento.config.state.setMessage();
```
Frodigo, line 37: there is no method removeMessage? It's not clear to me that `setMessage()` clears cookie.

packages/theme/composables/useMagentoConfiguration.ts
```diff
@@ -37,25 +36,19 @@ export const useMagentoConfiguration: UseMagentoConfiguration = () => {
 
     // eslint-disable-next-line promise/catch-or-return
     loadConfig().then(() => {
-      if (!app.$cookies.get(cookieNames.storeCookieName) || updateCookies) {
-        app.$cookies.set(
-          cookieNames.storeCookieName,
-          storeConfigGetters.getCode(storeConfig.value as StoreConfig),
-        );
+      if (!app.$vsf.$magento.config.state.getStore() || updateCookies) {
+        // @ts-ignore
```
sethidden, line 40: I added generics now so I could remove the ts-ignores and leave a comment that those types aren't entirely correct

packages/theme/composables/useUiNotification/index.ts
sethidden: @Frodigo Hmm, I recently moved the setMessage getMessage and removeMessage methods to one container so now it's config.state.message.set .get and .remove, but that makes it necessary to introduce breaking changes into magento-api.

Is it ok to keep the old approach of having setMessage, getMessage and removeMessage on the top level? So ~18 methods in one container. It's pretty verbose and not very fun but I'd like to avoid messing a lot of magento-api types to reduce the damage of migrating from rc6 to rc7 for example.

packages/theme/composables/useMagentoConfiguration.ts
sethidden: nope

packages/theme/composables/useUiNotification/index.ts
sethidden: /\ Done what I said above

packages/theme/composables/useMagentoConfiguration.ts
```diff
@@ -37,25 +36,19 @@ export const useMagentoConfiguration: UseMagentoConfiguration = () => {
 
     // eslint-disable-next-line promise/catch-or-return
     loadConfig().then(() => {
-      if (!app.$cookies.get(cookieNames.storeCookieName) || updateCookies) {
-        app.$cookies.set(
-          cookieNames.storeCookieName,
-          storeConfigGetters.getCode(storeConfig.value as StoreConfig),
-        );
+      if (!app.$vsf.$magento.config.state.getStore() || updateCookies) {
```
sethidden, line 39: There's something wrong with the 3 ifs here. They're making the app send serialized objects inside cookies instead of regular string values

sethidden: Facepalm i removed the getters

Frodigo:
@sethidden please provide a description for this pull request.

sethidden:
@Frodigo is this ok?

packages/theme/plugins/token-expired.ts
```diff
@@ -20,25 +22,26 @@ const hasAuthorizationError = (res): boolean => {
...
+      id: Symbol(''),
+      message: app.i18n.t('You are not authorized, please log in.') as string,
       type: 'warning',
       icon: null,
       persist: true,
       title: null,
     });
 
-    app.router.go(app.localePath('/'));
+    app.router.push(app.localePath('/'));
```
Frodigo, line 41: Can you explain this change?

Frodigo (changes requested):

Frodigo:
> @Frodigo is this ok?

yes, thanks!

sethidden:
https://router.vuejs.org/api/#go <https://router.vuejs.org/api/#go> go takes a number as argument, not a string

sethidden:
(Sending again because email responses to gh notifications got broken)
https://router.vuejs.org/api/#go <https://router.vuejs.org/api/#go> takes a number as argument, not a string

packages/theme/plugins/token-expired.ts
sethidden: https://router.vuejs.org/api/#go takes a number as argument, not string

Frodigo: Thanks!

Frodigo (approved):

PR #812, opened 2022-04-04
M2 320 sanitize html server side to reduce client bundle size

sethidden:
We currently sanitize CmsContent.vue both server and client side. It can be moved to server middleware to sanitize only server-side and save ~60kb on the client (non gzipped) bundle

- [ ] Remove require('fs') from config.js as part of #813


Frodigo:
@sethidden I am closing this one. Feel free to re-open it when you will have a chance to back into it.

PR #821, opened 2022-04-07
docs: add JSDoc comments to useAddress composable, its types and the related API methods

VitorLuizC:
## Description
- Change `useAddresses` composable and the API methods `createCustomerAddress`, `deleteCustomerAddress`, `getCustomerAddresses`, and `updateCustomerAddress` to **function declarations**.

- Change `UseAddressesErrors` and `UseAddressesInterface` types to **interfaces**.

- Add JSDoc comments to `useAddresses` composable, its interfaces, and the related API methods.

## Related Issue

https://vsf.atlassian.net/browse/M2-372 ?

## Motivation and Context

Documents the `useAddresses` with examples for the community.

## Types of changes

## Checklist:


Frodigo (approved):

packages/theme/composables/useAddresses/index.ts
```diff
@@ -1,10 +1,86 @@
...
+ *
+ * - {@link @vue-storefront/magento-api#createCustomerAddress} for saving new customer address;
+ *
+ * - {@link @vue-storefront/magento-api#updateCustomerAddress} for updating existing customer address;
+ *
+ * - {@link @vue-storefront/magento-api#deleteCustomerAddress} for removing existing customer address;
+ *
+ * It is currently used in:
+ *
+ * - `pages/MyAccount/AddressesDetails.vue`
```
sethidden, line 32: @Frodigo is this a convention we have to mention that? If you want to find that out, you can use "Go to references" in an IDE. It's going to be difficult to maintain down the road

packages/theme/composables/useAddresses/index.ts
```diff
@@ -1,10 +1,86 @@
 import { Ref, ref, useContext } from '@nuxtjs/composition-api';
 import { Logger } from '~/helpers/logger';
 import { transformUserCreateAddressInput, transformUserUpdateAddressInput } from '~/helpers/userAddressManipulator';
-import { ComposableFunctionArgs } from '~/composables/types';
-import { UseAddressesInterface, UseAddressesParamsInput, UseAddressesErrors } from '~/composables/useAddresses/useAddresses';
+import type { ComposableFunctionArgs } from '~/composables/types';
+import type { UseAddressesInterface, UseAddressesParamsInput, UseAddressesErrors } from './useAddresses';
 
-export const useAddresses = () : UseAddressesInterface => {
+/**
```
sethidden, line 7: There are 2 types of addresses - billing and shipping. Could you please add some info about how are they related to this composable? Does it return both types at once or does it need some searchParams switch?

packages/theme/composables/useAddresses/index.ts
```diff
@@ -109,6 +185,8 @@ export const useAddresses = () : UseAddressesInterface => {
     update,
     remove,
   };
-};
+}
+
+export * from './useAddresses';
```
sethidden, line 190: Doesn't `export *` work aswell? Why the change  https://github.com/vuestorefront/magento2/pull/821/files#diff-0e5009483ab73095c0fc0d47343ff91c29aaf90ebe2034d3f7ed03919185bb41R34 aswell? Is this related to api-extractor's detection?

Frodigo: Yes it's related to api-extractor. I tested it before and it's necessary to export all types fo build docs.

packages/theme/composables/useAddresses/index.ts
Frodigo: but in Magento address object is only one. There is no billing address type or shipping address type.

packages/theme/composables/useAddresses/index.ts
Frodigo: It's useful if you read the documentation as a Web page.

Frodigo:
Hi, @VitorLuizC changes that you requested about useCategory have been merged to the develop branch. Could you update this PR to cover that changes?

packages/theme/composables/useAddresses/index.ts
VitorLuizC: @sethidden agree. It will be hard to keep these links updated, mainly because updating the components and pages gave no clue that contributors should also update composable docs.

@Frodigo maybe it would be better if I remove this section. What do you think?

packages/theme/composables/useAddresses/index.ts
VitorLuizC: > Doesn't export * work aswell?

As @Frodigo said, the api-extractor requires exporting the types. And ESLint ([`unicorn/prefer-export-from`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v40.1.0/docs/rules/prefer-export-from.md)) doesn't allow exporting the same types imported at the beginning of the file, so I have to use `export *`.

> Why the change https://github.com/vuestorefront/magento2/pull/821/files#diff-0e5009483ab73095c0fc0d47343ff91c29aaf90ebe2034d3f7ed03919185bb41R34 aswell? Is this related to api-extractor's detection?

You're right if you're talking about the paths. The api-extractor wasn't working with the `~` alias. But if you're talking about `import type`, I changed it because it's better for bundlers and transpilers, so they don't need to require modules only used because of imported types.

packages/theme/composables/useAddresses/index.ts
VitorLuizC: I'll say in the docs that the addresses are used for both billing and shipping.

Frodigo (approved):

PR #839, opened 2022-04-12
refactor: move graphql client to context

ghost:
this is so you can query graphql endpoints in plugins

The getHeaderProxy(ctx) may seem a little convoluted but:
- graphql-request doesn't support dynamic headers https://github.com/prisma-labs/graphql-request/pull/91
- it's possible to just dynamically calculate getHeaders() before each request in useApi (just add a third argument to "request"), but that way if someone wants to query something from the a Nuxt plugin they'd need to reimplement the cookie logic

This PR also makes sure that the clients are instantiated only once - before each new request had a new client

### DIFF
--- a/packages/theme/composables/useApi/index.ts
+++ b/packages/theme/composables/useApi/index.ts
@@ -1,54 +1,14 @@
 import { useContext } from '@nuxtjs/composition-api';
-import { request as graphQLRequest, GraphQLClient } from 'graphql-request';
+import type { Variables } from 'graphql-request';
+
+type Request = <T = any, V = Variables>(document: string, variables?: V, requestHeaders?: HeadersInit) => Promise<T>;
 
 export const useApi = () => {
-  const { app } = useContext();
-  const customerToken = app.$vsf.$magento.config.state.getCustomerToken();
-  const storeCode = app.$vsf.$magento.config.state.getStore();
-  const currency = app.$vsf.$magento.config.state.getCurrency();
-  const magentoConfig = app.$vsf.$magento.config;
-  // TODO remove once we remove apollo client
-  const { useGETForQueries } = magentoConfig.customApolloHttpLinkOptions;
-  const defaultEndpoint = magentoConfig.magentoApiEndpoint;
-  const defaultHeaders: {
-    authorization?: string,
-    store?: string
-    'Content-Currency'?: string
-  } = {};
+  const context = useContext();
+  const query : Request = (document, variables, headers) => context.$graphql.query.request(document, variables, headers);
+  const mutate : Request = (document, variables, headers) => context.$graphql.mutation.request(document, variables, headers);
 
-  if (customerToken) {
-    defaultHeaders.authorization = `Bearer ${customerToken}`;
-  }
-
-  if (storeCode) {
-    defaultHeaders.store = storeCode;
-  }
-
-  if (currency) {
-    defaultHeaders['Content-Currency'] = currency;
-  }
-
-  const query = (
-    document: string,
-    variables = null,
-    endpoint = defaultEndpoint,
-    requestHeaders = defaultHeaders,
-  ) => new GraphQLClient(endpoint, {
-    method: useGETForQueries ? 'GET' : 'POST',
-    headers: requestHeaders,
-  }).request(document, variables);
-
-  const mutate = (
-    document: string,
-    variables = null,
-    endpoint = defaultEndpoint,
-    requestHeaders = defaultHeaders,
-  ) => graphQLRequest(endpoint, document, variables, requestHeaders);
-
-  return {
-    query,
-    mutate,
-  };
+  return { query, mutate };
 };
 
 export default useApi;
--- a/packages/theme/nuxt.config.js
+++ b/packages/theme/nuxt.config.js
@@ -212,6 +212,7 @@ export default () => {
       '~/plugins/i18n',
       '~/plugins/fcPlugin',
       '~/plugins/dompurify',
+      '~/plugins/graphqlClient',
     ],
     serverMiddleware: [
       '~/serverMiddleware/body-parser.js',
@@ -225,6 +226,9 @@ export default () => {
     },
     image: {
       provider: process.env.VSF_IMAGE_PROVIDER,
+    },
+    env: {
+      VSF_MAGENTO_GRAPHQL_URL: process.env.VSF_MAGENTO_GRAPHQL_URL,
     },
   };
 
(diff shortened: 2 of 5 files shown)


packages/theme/plugins/graphqlClient.ts
```diff
@@ -0,0 +1,84 @@
...
+    },
+  };
+
+  return new Proxy({}, proxyHandler);
+};
+
+type GraphQlClients = Record<'query' | 'mutation', GraphQLClient>;
+
+const plugin : Plugin = (ctx, inject) => {
+  // TODO remember to remove the magento http options aswell
```
sethidden, line 61: this

sethidden:
@Frodigo do you still want me to mark the useApi composable as deprecated? I'd probably need to make a task to replace all the ocurrences

packages/theme/composables/useApi/index.ts
```diff
@@ -1,54 +1,14 @@
...
 
-export const useApi = () => {
-  const { app } = useContext();
-  const customerToken = app.$vsf.$magento.config.state.getCustomerToken();
-  const storeCode = app.$vsf.$magento.config.state.getStore();
-  const currency = app.$vsf.$magento.config.state.getCurrency();
-  const magentoConfig = app.$vsf.$magento.config;
-  // TODO remove once we remove apollo client
-  const { useGETForQueries } = magentoConfig.customApolloHttpLinkOptions;
-  const defaultEndpoint = magentoConfig.magentoApiEndpoint;
```
sethidden, line 12: Hmm magentoApiEndpoint also seems like a candidate to be removed completely from the codebase (we have env var)

docs/guide/graphql-get.md
```diff
@@ -45,10 +45,7 @@ sub process_graphql_headers {
 
 You can find more information about this topic at [https://devdocs.magento.com/guides/v2.4/graphql/caching.html#caching-with-varnish]
 
-3. Open the `middleware.config.js` file, it should be located on the root folder of your store, there you will change the flag `useGETForQueries` to `true` to enable the usage of GET request for queries
-
+3. Make sure your GraphQL client sends queries (but not mutations) using the HTTP GET method. Configuring this for `graphql-request` is as follows:
```
Frodigo, line 48: I think we shouldn't remove that point 3 because it's still valid and it's about server middleware. 

There should be another document that describes the new GraphQL, and actually, we have a ticket in the backlog for that, so let's restore this change, please.

packages/api-client/src/types/setup.ts
```diff
@@ -92,7 +91,6 @@ export interface Config<T = any> extends ClientConfig {
   client?: ApolloClient<T>;
   storage: Storage;
   customOptions?: ApolloClientOptions<any>;
-  customApolloHttpLinkOptions?: HttpOptions;
```
Frodigo, line 95: Related to the previous comment, please restore this change.

packages/api-client/src/helpers/magentoLink/graphQl.ts
```diff
@@ -67,7 +67,6 @@ export const apolloLinkFactory = (settings: Config, handlers?: {
     fetchOptions: {
       agent,
     },
-    ...settings.customApolloHttpLinkOptions,
```
Frodigo, line 70: Related to the previous comment, please restore this change.

packages/theme/middleware.config.js
```diff
@@ -21,9 +21,6 @@ module.exports = {
             default: isCheckoutEnabled,
           },
         },
-        customApolloHttpLinkOptions: {
```
Frodigo, line 24: As I mentioned before we won't remove this.

docs/guide/graphql-get.md
Frodigo: Som in the new API client we won't use that `useGetForQueries` config option, but documentation will be done in separate PR.

packages/theme/plugins/graphqlClient.ts
```diff
@@ -0,0 +1,84 @@
+import { GraphQLClient } from 'graphql-request';
+import { Plugin, Context } from '@nuxt/types';
```
Frodigo, line 2:
```suggestion
import type { Plugin, Context } from '@nuxt/types';
```

packages/theme/plugins/graphqlClient.ts
```diff
@@ -0,0 +1,84 @@
...
+        enumerable: true,
+        configurable: true,
+      };
+    },
+  };
+
+  return new Proxy({}, proxyHandler);
+};
+
+type GraphQlClients = Record<'query' | 'mutation', GraphQLClient>;
```
Frodigo, line 58: The name of this is weird to me. WHat do you think about `GraphQlOperations`?

packages/theme/nuxt.config.js
```diff
@@ -226,6 +225,9 @@ export default () => {
     image: {
       provider: process.env.VSF_IMAGE_PROVIDER,
     },
+    env: {
```
Frodigo, line 228: Why this is needed here? I  see in the code that you use `process.env.VSF_MAGENTO_GRAPHQL_URL;` so I don't understand the purpose of this config.

RetroModernDev (changes requested):

packages/theme/nuxt.config.js
sethidden: Environment variables that are passed to nuxt are only available server side by default. So if I tried to go somewhere client-side the requests would fail because their VSF_MAGENTO_GRAPHQL_URL would be undefined.

This works without useApi beacuse it uses the middleware.config.js magentoApiEndpoint https://github.com/vuestorefront/magento2/pull/839/files#diff-66e4ffc399d4ad1f1464d30529bbb1ff498032d4293f1b3e83fdb33d4ee42e6dL12

Since you said you prefer not to modify api-config things (so I can't remove magentoApiEndpoint) I could either:
1. remove https://github.com/vuestorefront/magento2/pull/839/files#diff-ceda593cdb4e231ccc447030a0fd70207f4b6d2f7cbd5e6cff74d1309270c5e7R62 and keep using magentoApiEndpoint
2. keep using VSF_MAGENTO_GRAPHQL_URL directly since it's one layer less

what do you think?

Frodigo: I see, let's go with point 2. keep using VSF_MAGENTO_GRAPHQL_URL directly since it's one layer less

packages/theme/plugins/graphqlClient.ts
sethidden: Well, they aren't operations. They're instances of a client that can make requests. This is a different approach from what useApi used which created a new client for each request. So the approach here is "upside down" as to what was with useApi

sethidden: i.e. you still need to do ctx.$graphql.query.request(), it's not query()

Frodigo: Ok, I see, thanks, I understand now so no changes are required.

RetroModernDev (approved):

PR #855, opened 2022-04-13
refactor: improve pinia typing

ghost:
1. Make `this.` type inside pinia stores actually return types it's supposed to: state & getters (instead of being just "any")

2. Enable typings for @pinia/nuxt then use the pinia nuxt plugin - so now the "pass api as argument to store" is not needed anymore thans to the nuxt plugin. Graphql client is now in the nuxt context so it's easy to just this.$nuxt.$graphql.query.request

For some reason the this.$nuxt property has proper typings only in tsserver, but with Volar's take over mode it doesn't work ($nuxt does not exist no this). But this seems like a Volar issue to me as Volar needs extra config for Vue 2 which is not present in this repo yet (will do later)

### DIFF
--- a/packages/theme/modules/magento/index.ts
+++ b/packages/theme/modules/magento/index.ts
@@ -1,8 +1,9 @@
+import { Module } from '@nuxt/types';
+
 /* eslint-disable unicorn/prefer-module */
 const path = require('path');
 
-// eslint-disable-next-line func-names
-export default function (options) {
+const nuxtModule : Module = function (options) {
   this.extendBuild((config) => {
     // eslint-disable-next-line no-param-reassign
     config.resolve.alias['@vue-storefront/magento-api$'] = require.resolve('@vue-storefront/magento-api');
@@ -12,4 +13,6 @@ export default function (options) {
     src: path.resolve(__dirname, './plugin.ts'),
     options,
   });
-}
+};
+
+export default nuxtModule;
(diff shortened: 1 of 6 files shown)


packages/theme/modules/magento/index.ts
```diff
@@ -1,8 +1,9 @@
+import { Module } from '@nuxt/types';
+
 /* eslint-disable unicorn/prefer-module */
 const path = require('path');
 
-// eslint-disable-next-line func-names
-export default function (options) {
+const nuxtModule : Module = function (options) {
```
sethidden, line 6: the noImplicitAny I added in tsconfig.json puts some stricter rules on what "this" means in code, so I had to add the `: Module` type assertion to make TS know what "this" means (before this was any)

packages/theme/stores/category.ts
```diff
@@ -1,21 +1,19 @@
...
 
-export const useCategoryStore = (api: ReturnType<typeof useApi>, $pinia?: Pinia) => defineStore('category', {
+export const useCategoryStore = defineStore('category', {
   state: (): CategoryState => ({
     rawCategories: null,
   }),
   actions: {
     async load() {
-      this.rawCategories = await api.query(categoryListGql);
+      this.rawCategories = await this.$nuxt.$graphql.query.request(categoryListGql);
```
sethidden, line 16: yay

RetroModernDev (approved):

PR #857, opened 2022-04-14
refactor: refactored useCurrency `load` and `change` methods

Frodigo:
## Description
- The change method of useCurrency is synchronous so it's not necessary to set the loading state there.
- customQuery was not passed correctly to API, and it's fixed, params type also is fixed

### DIFF
--- a/packages/theme/composables/useCurrency/index.ts
+++ b/packages/theme/composables/useCurrency/index.ts
@@ -11,7 +11,7 @@ const useCurrency = (): UseCurrency => {
   const configStore = useConfigStore();
   const currency = computed(() => configStore.currency);
 
-  const load = async (params?: ComposableFunctionArgs<{ currency: string }>) => {
+  const load = async (params?: ComposableFunctionArgs<{}>) => {
     error.value.load = null;
     loading.value = true;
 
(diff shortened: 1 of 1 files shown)


packages/theme/composables/useCurrency/index.ts
```diff
@@ -32,8 +32,6 @@ const useCurrency = (): UseCurrency => {
 
   const change = (params: ComposableFunctionArgs<{ id: string }>) => {
```
VitorLuizC, line 33: Since it doesn't support custom queries, we can use just the literal type here.

```suggestion
  const change = (params: { id: string }) => {
```

VitorLuizC (changes requested):

packages/theme/composables/useCurrency/index.ts
```diff
@@ -11,14 +11,14 @@ const useCurrency = (): UseCurrency => {
   const configStore = useConfigStore();
   const currency = computed(() => configStore.currency);
 
-  const load = async (params?: ComposableFunctionArgs<CustomQuery>) => {
+  const load = async (params?: ComposableFunctionArgs<{ currency: string }>) => {
```
VitorLuizC, line 14: The type received by `ComposableFunctionArgs` is the custom query, but something to combine with the object that has custom query.

`ComposableFunctionArgs<{ currency: string }>` became `{ customQuery?: CustomQuery, currency: string }` instead of `{ customQuery?: { currency: string } }`, and since we're not using `currency` in the method I think we can just pass an empty object instead.

```suggestion
  const load = async (params?: ComposableFunctionArgs<{}>) => {
```

VitorLuizC (changes requested):

VitorLuizC (approved):
LGTM

PR #859, opened 2022-04-14
perf!: load whole user object only when necessary

sethidden:
wait for: https://github.com/vuestorefront/magento2/pull/862

Currently the user/customer object is pretty big and is excessive for checking if the user is logged in. A simple boolean should be sufficient.

Now instead of loading the user in the default layout, the login status is calculated outside of vue in the token-expired plugin at the start of the application

Also fixed the logic in MyAccount.vue where it'd excessively query for the user/customer object

- [ ] fix conflicts
- [ ] mark breaking change for login

### DIFF
--- a/packages/theme/composables/useUser/index.ts
+++ b/packages/theme/composables/useUser/index.ts
@@ -13,7 +13,7 @@ export const useUser = (): UseUser => {
   const { app } = useContext();
   const { setCart } = useCart();
   const loading: Ref<boolean> = ref(false);
-  const isAuthenticated = computed(() => Boolean(customerStore.user?.firstname));
+  const isAuthenticated = computed(() => customerStore.isLoggedIn);
   const errorsFactory = () => ({
     updateUser: null,
     register: null,
@@ -88,9 +88,10 @@ export const useUser = (): UseUser => {
 
       await app.context.$vsf.$magento.api.revokeCustomerToken({ customQuery });
 
-      apiState.setCustomerToken(null);
-      apiState.setCartId(null);
+      apiState.removeCustomerToken();
+      apiState.removeCartId();
       setCart(null);
+      customerStore.setIsLoggedIn(false);
       error.value.logout = null;
       customerStore.user = null;
     } catch (err) {
@@ -133,7 +134,7 @@ export const useUser = (): UseUser => {
   };
 
   // eslint-disable-next-line @typescript-eslint/require-await,no-empty-pattern
-  const login = async ({ user: providedUser, customQuery }) => {
+  const login = async ({ user: providedUser, customQuery }) : Promise<void> => {
     Logger.debug('[Magento] useUser.login', providedUser);
     resetErrorValue();
 
@@ -159,6 +160,7 @@ export const useUser = (): UseUser => {
       if (!data.generateCustomerToken || !data.generateCustomerToken.token) {
         Logger.error('Customer sign-in error'); // todo: handle errors in better way
       }
+      customerStore.setIsLoggedIn(true);
       apiState.setCustomerToken(data.generateCustomerToken.token);
 
       // merge existing cart with customer cart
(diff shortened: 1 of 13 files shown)


sethidden:
For some reason the tiny __typename fetch also returns orders?

packages/theme/pages/MyAccount.vue
```diff
@@ -5,6 +5,7 @@
       :breadcrumbs="breadcrumbs"
     />
     <SfContentPages
+      v-if="user"
```
sethidden, line 8: probably should use some skeleton loader instead of causing 100vh of layout shift

packages/theme/layouts/default.vue
```diff
@@ -23,7 +23,7 @@ import LazyHydrate from 'vue-lazy-hydration';
 import {
```
sethidden, line 23: https://github.com/vuestorefront/magento2/pull/862
This will conflict with that PR, so I'll be a gentleman and let Bart's PR pass first

packages/theme/pages/MyAccount.vue
```diff
@@ -13,28 +13,52 @@
     >
       <SfContentCategory :title="$t('Personal Details')">
         <SfContentPage :title="$t('My profile')">
-          <MyProfile />
+          <SkeletonLoader
+            v-if="$fetchState.pending"
```
sethidden, line 17: As you know, SfContentPage just uses display-none instead of v-if undreneath, so I need to hard-block rendering using v-if here to allow "user is undefined" errors before fetch loads user data

packages/theme/plugins/token-expired.ts
```diff
@@ -1,32 +1,25 @@
...
+import loginStatusPingQueryGql from '~/composables/useUser/loginStatusPingQuery.gql';
+import { useCustomerStore } from '~/stores/customer';
 
-const hasAuthorizationError = (res: AxiosResponse): boolean => {
-  if (!res?.data?.errors) {
-    return false;
-  }
-
-  const { errors } = res.data;
+export const hasGraphqlAuthorizationError = (res: ApolloQueryResult<unknown>) => res?.errors
```
sethidden, line 7: refactored imperative hasAuthorizationError to more declarative

packages/theme/plugins/token-expired.ts
```diff
@@ -1,32 +1,25 @@
...
+const plugin : Plugin = async ({ $pinia, app }) => {
+  const customerStore = useCustomerStore($pinia);
 
-  // eslint-disable-next-line no-restricted-syntax
-  for (const error of errors) {
-    if (error?.extensions?.category === 'graphql-authorization') {
-      isAuthErr = true;
-      break;
-    }
+  const responseOfLoginStatusPing = await app.$vsf.$magento.api.customQuery({ query: loginStatusPingQueryGql });
```
sethidden, line 13: initial ping to check if the user is logged in at the start of the app. thanks to this, the default layout contains one less critical query (and after Bart's PR, there will be none)

packages/theme/stores/customer.ts
```diff
@@ -1,22 +1,30 @@
...
 }
 
 export const useCustomerStore = defineStore('customer', {
   state: (): CustomerState => ({
     wishlist,
     user,
     cart,
+    isLoggedIn: false,
   }),
+  actions: {
```
sethidden, line 25: This may seem excessive as you can just do customerStore.isLoggedIn to modify the state directly, but I prefer to add an action to define an API that defines what operations can be done on the store

packages/theme/composables/useUser/index.ts
```diff
@@ -185,19 +187,16 @@ export const useUser = (): UseUser => {
       }
 
       error.value.login = null;
-      customerStore.user = load();
```
sethidden, line 188: logging in no longer fetches the user object, it just sets the tiny login status boolean to true

Frodigo: I see, so mark this as a breaking change in the PR.

Frodigo (dismissed):
Good job, thanks!

packages/theme/composables/useUser/index.ts
sethidden: Added `!` to commit conventional-commits prefix

Frodigo (approved):

PR #862, opened 2022-04-14
refactor: store config query

bartoszherba:
## Description
- move loading of the store config to the plugin
- add store config documentation

## Motivation and Context
Perf improvement and better code base

## How Has This Been Tested?
-

## Types of changes

## Checklist:

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -51,21 +51,6 @@ This type contains information about the Magento's Store Configuration which is 
 
 Plugin `plugins/storeConfigPlugin.ts` is responsible for initial population of the Store Config data based on query in `plugins/query/StoreConfig.gql.ts`. To modify the initial Store Configuration state simply adjust the query to your needs.
 
-```plain
-query storeConfig {
-    storeConfig {
-        store_code,
-        default_title,
-        store_name,
-        default_display_currency_code,
-        locale,
-        header_logo_src,
-        logo_width,
-        logo_height,
-        logo_alt
-    }
-  }
-```
 ## Multistore and localization
 
 Each Magento Store View need to have corresponding locale configuration object in `i18n.locales` array in `nuxt.config.js`.
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -45,6 +45,27 @@ Then on the `config` folder create a file `dev.json` with your configurations.
 }
 ```
 
+
+## Store Config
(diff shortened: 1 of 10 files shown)


packages/theme/plugins/storeConfigPlugin.ts
```diff
@@ -0,0 +1,23 @@
+import { PiniaPluginContext } from 'pinia';
+import { Plugin } from '@nuxt/types';
+import { ref, set } from '@nuxtjs/composition-api';
+import StoreConfigGql from '~/plugins/query/StoreConfig.gql';
+
+const storeConfigPlugin: Plugin = async ({ $pinia, $graphql }) => {
+  const configData = await $graphql.query.request(StoreConfigGql);
+  $pinia.use(({ store }: PiniaPluginContext) => {
+    if (store.$id !== 'magentoConfig') return;
+    // eslint-disable-next-line no-prototype-builtins
```
sethidden, line 10:
```suggestion
```
There are no prototype builtins used on the following line, so this eslint magic comment doesn't do anything

packages/theme/components/StoreSwitcher/StoresModal.vue
```diff
@@ -62,7 +62,10 @@ export default defineComponent({
   },
   props: {
     isLangModalOpen: Boolean,
-    storeConfig: Object,
+    storeConfig: {
+      type: Object as PropType<StoreConfig>,
+      default: () => {},
```
sethidden, line 67: eh we have the annoying eslint rule where it forces us to provide a default value

would it be possible to use like as `PropType<StoreCofnig | null>`? The `{}` doesn't really match StoreConfig

docs/guide/configuration.md
```diff
@@ -45,6 +45,27 @@ Then on the `config` folder create a file `dev.json` with your configurations.
...
+        store_name,
+        default_display_currency_code,
+        locale,
+        header_logo_src,
+        logo_width,
+        logo_height,
+        logo_alt
+    }
+  }
+```
```
sethidden, line 68:
```suggestion
```
This is already in StoreConfig.gql.ts and the added docs contents already point to that file - do we want to maintain this here aswell?

Frodigo (dismissed):

packages/theme/components/StoreSwitcher/StoresModal.vue
bartoszherba: Maybe we should remove that rule then?

bartoszherba: Anyway, I updated the code.

sethidden: Did you push it out? I can't see the updates

sethidden (approved):

PR #873, opened 2022-04-15
ci: github tokens don't launch deployment_status events

ghost:
Github doesn't allow triggering actions from actions if the default github access token is used.
So in our case, it doesn't allow triggering the "Run SpeedCurve test" action from the "deploy to vsf cloud" action
I generated my own for my own account just with the repo_deployments permission. My token will expire in 30 days

This token is just for pinging the github api of deployments, which is basically a CRUD for a list of objects. This isn't some critical stuff like SSH keys or anything.

https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

### DIFF
--- a/.github/workflows/deployment-template.yml
+++ b/.github/workflows/deployment-template.yml
@@ -18,6 +18,8 @@ on:
          required: true
        cloud_password:
          required: true
+       deployment_status_token: # can be ${{ github.token }}
+         required: true
 
 
 jobs:
@@ -28,7 +30,7 @@ jobs:
         name: Create GitHub deployment
         id: deployment
         with:
-          token: ${{ github.token }}
+          token: ${{ secrets.deployment_status_token }}
           environment: ${{ inputs.github_environment_name }}
           initial_status: in_progress
           target_url: ${{ inputs.target_url }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deployment-template.yml
```diff
@@ -28,7 +30,7 @@ jobs:
         name: Create GitHub deployment
         id: deployment
         with:
-          token: ${{ github.token }}
+          token: ${{secrets.deployment_status_token }}
```
inkoguto, line 33: pls reformat (lack of space after "{{")

.github/workflows/deployment-template.yml
```diff
@@ -59,7 +61,7 @@ jobs:
         if: failure()
         uses: chrnorm/deployment-status@releases/v1
         with:
-          token: ${{ github.token }}
+          token: ${{secrets.deployment_status_token }}
```
inkoguto, line 64: pls reformat (lack of space after "{{")

inkoguto (dismissed):

inkoguto (approved):

PR #878, opened 2022-04-18
feat!: make `error` and `loading` properties readonly on `useMakeOrder` composable 

VitorLuizC:
## Description
- **[BREAKING CHANGE]** Make the `error` and `loading` properties readonly (they're refs);
- Create `UseMakeOrderInterface` similar as other composables;
- Write TSDocs for `placeOrder` API method and its GraphQL query;
- Write TSDocs for `useMakeOrder` composable and its types;
- Convert `useMakeOrder`, `placeOrder` into function declarations (since it's better for the docs generator);
- Fix `command` used by `useMakeOrder`;
- Export `useMakeOrder` types in `~/composables`.

## Related Issue
https://vsf.atlassian.net/browse/M2-387

## Motivation and Context
Improves the DX of the `useMakeOrder` composable.

## Types of changes

## Checklist:


packages/theme/composables/useMakeOrder/commands/placeOrderCommand.ts
```diff
@@ -1,9 +1,12 @@
...
+type Context = ReturnType<typeof useContext>;
 
 export const placeOrderCommand = {
-  execute: async (context, cartId: string): Promise<PlaceOrderOutput> => {
-    const { data } = await context.$vsf.$magento.api.placeOrder({ cart_id: cartId });
+  execute: async (context: Context, cartId: string): Promise<PlaceOrderOutput> => {
+    const { data } = await context.app.$vsf.$magento.api.placeOrder({ cart_id: cartId });
 
-    return data.placeOrder.order;
+    return data.placeOrder;
```
VitorLuizC, line 10: I found a way to type the `context`, but TS started to complain when I did. First, because `$vsf` doesn't exist in `context` but in `context.app`; and then because `data.placeOrder.order` is of type `Order` instead of `PlaceOrderOutput` which is its parent (`data.placeOrder`).

It looks like I fixed an error in `placeOrderCommand`, but I would appreciate it if you guys review this part with care.

sethidden: You can also just use import { Context } from '@nuxt/types' since that's the underyling type of ReturnType<useContext>

packages/theme/composables/useMakeOrder/commands/placeOrderCommand.ts
```diff
@@ -1,9 +1,10 @@
-import { PlaceOrderOutput } from '~/modules/GraphQL/types';
+import type { NuxtAppOptions } from '@nuxt/types';
+import type { PlaceOrderOutput } from '~/modules/GraphQL/types';
 
 export const placeOrderCommand = {
-  execute: async (context, cartId: string): Promise<PlaceOrderOutput> => {
-    const { data } = await context.$vsf.$magento.api.placeOrder({ cart_id: cartId });
+  execute: async (app: NuxtAppOptions, cartId: string): Promise<PlaceOrderOutput | null> => {
```
sethidden, line 5:
```suggestion
  execute: async (context: Context, cartId: string): Promise<PlaceOrderOutput | null> => {
```

All our other *Command.ts files use context as the arg here, so even if the @nuxt/types Context doesn't work, the not-so-beautiful Return<useContext> type is better than changing the convention

packages/theme/composables/useMakeOrder/commands/placeOrderCommand.ts
sethidden: Followed up in https://github.com/vuestorefront/magento2/pull/878#discussion_r853762397

packages/theme/composables/useMakeOrder/index.ts
```diff
@@ -1,22 +1,23 @@
...
+  const { app } = useContext();
 
-  const make = async (): Promise<Maybe<PlaceOrderOutput>> => {
+  const make = async (): Promise<PlaceOrderOutput | null> => {
     Logger.debug('useMakeOrder.make');
     let placedOrder = null;
     try {
       loading.value = true;
-      placedOrder = await placeOrderCommand.execute(context, cart.value.id);
+      placedOrder = await placeOrderCommand.execute(app, cart.value.id);
```
Frodigo, line 20: Context must be pass here, not app.

Frodigo (changes requested):

Frodigo (approved):

PR #887, opened 2022-04-19
feat!: make `error` and `loading` properties readonly on `useUser` composable

filipsobol:
This PR updates the `useUser` composable:

- [BREAKING CHANGE] wraps the `loading` and `error` objects with `readonly` so they cannot be manipulated outside of the composable,
- [BREAKING CHANGE] renames `UseUser` interface to `UseUserInterface` for consistency with other composables,
- improves typing,
- adds a short description to all interfaces, types, and properties.


packages/theme/composables/useUser/useUser.ts
```diff
@@ -1,16 +1,19 @@
...
+  updateUser: Error | null;
+  register: Error | null;
+  login: Error | null;
+  logout: Error | null;
+  changePassword: Error | null;
+  load: Error | null;
 }
 
-export interface Customer {
+export interface User {
```
Frodigo, line 16: Changing the new type called User is confusing here. Magento returns the Customer type and we won't mix names (Customer and User) 

I recommend using the Customer type here returned from modules/graphql/types

Frodigo (changes requested):

packages/theme/composables/useUser/useUser.ts
filipsobol: This interface and the `Customer` interface generated from GraphQL and incompatible. I had to rename this to `User` because it collided with the `Customer` from GraphQL. I left some `TODO` comments for you to look into later

Frodigo: ok, thanks

Frodigo (approved):

PR #903, opened 2022-04-21
feat: header categories

bartoszherba:
## Description
- add multilevel categories navigation
- add unit tests for new components
- upgrade @testing-library/user-event @13 -> @14

## Motivation and Context
Improvement

Screenshot 2022-04-26 at 10 34 45

## Types of changes

## Checklist:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,13 @@ module.exports = {
   ],
   rules: {
     "@typescript-eslint/no-floating-promises": "off",
-    "jest/expect-expect": "off"
+    "jest/expect-expect": [
+      "error",
+      {
+        "assertFunctionNames": ["expect", "getByRole", "getByTestId"],
+      }
+    ],
+    "no-plusplus": "off",
   }
 }
+
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,6 +31,7 @@ module.exports = {
     '@vue-storefront/eslint-config-jest',
   ],
   rules: {
-    "@typescript-eslint/no-floating-promises": "off"
+    "@typescript-eslint/no-floating-promises": "off",
+    "jest/expect-expect": "off"
   }
 }
--- a/packages/theme/components/Header/Navigation/HeaderNavigation.vue
+++ b/packages/theme/components/Header/Navigation/HeaderNavigation.vue
@@ -1,0 +1,77 @@
+<template>
+  <div
+    class="header-navigation"
+  >
+    <div class="header-navigation__main">
+      <HeaderNavigationItem
+        v-for="(category, index) in categoryTree"
+        :key="index"
+        :data-testid="category.uid"
+        class="nav-item"
+        :label="category.name"
+        :link="localePath(getCatLink(category))"
+        @mouseenter.native="setCurrentCategory(category)"
+        @keyup.tab.native="setCurrentCategory(category)"
+      />
+    </div>
+    <HeaderNavigationSubcategories
+      v-if="hasChildren(currentCategory)"
+      :current-category="currentCategory"
+      @hideSubcategories="setCurrentCategory(null)"
+    />
+  </div>
+</template>
+<script lang="ts">
+import {
+  defineComponent, PropType, ref,
+} from '@nuxtjs/composition-api';
+import HeaderNavigationItem from './HeaderNavigationItem.vue';
+
+import { CategoryTree } from '~/modules/GraphQL/types';
+import { useUiHelpers } from '~/composables';
+
+export default defineComponent({
+  name: 'HeaderNavigation',
+  components: {
+    HeaderNavigationSubcategories: () => import('~/components/Header/Navigation/HeaderNavigationSubcategories.vue'),
+    HeaderNavigationItem,
+  },
+  props: {
+    categoryTree: {
+      type: Array as PropType<CategoryTree[]>,
+      default: () => [],
+    },
+  },
+  setup() {
+    const { getCatLink } = useUiHelpers();
+    const currentCategory = ref<CategoryTree>(null);
+
+    const setCurrentCategory = (category: CategoryTree | null) => {
+      currentCategory.value = category;
+    };
+
+    const hasChildren = (category: CategoryTree) => Boolean(category?.children);
+
+    return {
+      getCatLink,
+      setCurrentCategory,
+      currentCategory,
+      hasChildren,
+    };
+  },
+});
+</script>
+<style lang="scss" scoped>
+.header-navigation {
+  &__main {
+    display: flex;
+  }
+}
+.nav-item {
+  --header-navigation-item-margin: 0 var(--spacer-sm);
+
+  .sf-header-navigation-item__item--mobile {
+    display: none;
+  }
+}
+</style>
--- a/packages/theme/components/Header/Navigation/HeaderNavigation.vue
+++ b/packages/theme/components/Header/Navigation/HeaderNavigation.vue
@@ -1,23 +1,31 @@
 <template>
-  <div
-    class="header-navigation"
-  >
-    <div class="header-navigation__main">
+  <div class="header-navigation">
+    <div class="sf-header-navigation-item__item sf-header-navigation-item__item--desktop">
       <HeaderNavigationItem
         v-for="(category, index) in categoryTree"
         :key="index"
+        ref="lvl0CatRefs"
         :data-testid="category.uid"
-        class="nav-item"
         :label="category.name"
         :link="localePath(getCatLink(category))"
-        @mouseenter.native="setCurrentCategory(category)"
-        @keyup.tab.native="setCurrentCategory(category)"
+        tabindex="1"
+        aria-haspopup="true"
+        class="nav-item"
+        :data-index="index"
+        :has-children="hasChildren(category)"
+        @mouseenter.native.prevent="onMouseEnter(category)"
+        @keydown.down.native.prevent="setCurrentCategory(category)"
+        @keydown.up.native.prevent="setCurrentCategory(null)"
+        @keyup.tab.native.prevent="setFocus($event)"
+        @keydown.right.native.prevent="navRight()"
+        @keydown.left.native.prevent="navLeft()"
       />
     </div>
     <HeaderNavigationSubcategories
       v-if="hasChildren(currentCategory)"
       :current-category="currentCategory"
-      @hideSubcategories="setCurrentCategory(null)"
+      :has-focus="hasFocus"
+      @hideSubcategories="focusRootElementAndHideSubcategories"
     />
   </div>
 </template>
@@ -44,7 +52,12 @@ export default defineComponent({
   },
   setup() {
     const { getCatLink } = useUiHelpers();
+
     const currentCategory = ref<CategoryTree>(null);
+    const lvl0CatRefs = ref([]);
+    const hasFocus = ref(false);
+    let lvl0CatFocusIdx = 0;
+    let focusedElement = null;
 
     const setCurrentCategory = (category: CategoryTree | null) => {
       currentCategory.value = category;
@@ -52,11 +65,55 @@ export default defineComponent({
 
     const hasChildren = (category: CategoryTree) => Boolean(category?.children);
 
+    const setFocus = (event) => {
+      focusedElement = event.target;
+      lvl0CatFocusIdx = Number(event.target.dataset.index);
+      hasFocus.value = true;
+    };
+
+    const focusRootElementAndHideSubcategories = () => {
+      setCurrentCategory(null);
+      if (focusedElement !== null) focusedElement.focus();
+    };
+
+    const navRight = () => {
+      lvl0CatFocusIdx++;
+      if (lvl0CatRefs.value[lvl0CatFocusIdx]) {
+        lvl0CatRefs.value[lvl0CatFocusIdx].$el.focus();
+        focusedElement = lvl0CatRefs.value[lvl0CatFocusIdx].$el;
+      } else {
+        lvl0CatFocusIdx--;
+      }
+    };
+
+    const navLeft = () => {
+      lvl0CatFocusIdx--;
+      if (lvl0CatRefs.value[lvl0CatFocusIdx]) {
+        lvl0CatRefs.value[lvl0CatFocusIdx].$el.focus();
+        focusedElement = lvl0CatRefs.value[lvl0CatFocusIdx].$el;
+      } else {
+        lvl0CatFocusIdx++;
+      }
+    };
+
+    const onMouseEnter = (category: CategoryTree) => {
+      currentCategory.value = category;
+      focusedElement = null;
+      hasFocus.value = false;
+    };
+
     return {
       getCatLink,
       setCurrentCategory,
       currentCategory,
       hasChildren,
+      setFocus,
+      focusRootElementAndHideSubcategories,
+      lvl0CatRefs,
+      navRight,
+      navLeft,
+      hasFocus,
+      onMouseEnter,
     };
   },
 });
--- a/packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
+++ b/packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
@@ -1,0 +1,110 @@
+<template>
+  <div
+    v-if="hasChildren(currentCategory)"
+    data-testid="navigation-subcategories"
+    class="header-navigation__subcategories"
+  >
+    <div
+      v-for="(catLvl1, idxLvl1) in currentCategory.children"
+      :key="idxLvl1"
+      class="header-navigation__subcategory"
+    >
+      <SfLink
+        class="header-navigation__link1"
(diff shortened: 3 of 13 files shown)


.eslintrc.js
```diff
@@ -31,6 +31,7 @@ module.exports = {
     '@vue-storefront/eslint-config-jest',
   ],
   rules: {
-    "@typescript-eslint/no-floating-promises": "off"
+    "@typescript-eslint/no-floating-promises": "off",
+    "jest/expect-expect": "off"
```
sethidden, line 35:
```suggestion
```

You can't skip using "expect", even if getByRole returns an error

sethidden, line 35:
```suggestion
```

You can't skip using "expect", even if getByRole returns an error

packages/theme/components/Header/Navigation/__tests__/HeaderNavigation.spec.js
```diff
@@ -0,0 +1,50 @@
...
+import CategoryTreeDataMock from '~/test-utils/mocks/categoryTreeDataMock';
+
+jest.mock('~/composables');
+useUiHelpers.mockReturnValue(useUiHelpersMock());
+
+const sharedOptions = { props: { categoryTree: CategoryTreeDataMock[0].children } };
+describe('HeaderNavigation', () => {
+  it('displays main categories', () => {
+    const { getByRole } = render(HeaderNavigation, sharedOptions);
+    getByRole('link', { name: /women/i });
```
sethidden, line 16:
```suggestion
    expect(getByRole('link', { name: /women/i })).toBeTruthy();
```

sethidden, line 16:
```suggestion
    expect(getByRole('link', { name: /women/i })).toBeTruthy();
```

sethidden: (repeat for all the getByRoles below)

sethidden: (repeat for all the getByRoles below)

sethidden (commented):
Left some comments. Will leave some more in a second aswell because I'm scared this'll get merged while I'm reviewing again

sethidden (changes requested):
(meant to request changes)

packages/theme/components/Header/Navigation/__tests__/HeaderNavigationSubcategories.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    const { getByRole } = render(HeaderNavigationSubcategories, childrenProps);
+    getByRole('link', { name: /jackets/i });
+    getByRole('link', { name: /hoodies & sweatshirts/i });
+    getByRole('link', { name: /tees/i });
+    getByRole('link', { name: /tanks/i });
+  });
+
+  it('do not display any links if root category has no children', () => {
+    const { queryByTestId } = render(HeaderNavigationSubcategories, noChildrenProps);
+    expect(queryByTestId(/navigation-subcategories/i)).toBeNull();
```
sethidden, line 32:
```suggestion
    expect(queryByTestId('navigation-subcategories')).toBeNull();
```
As far as I've seen in the code you use `data-testid='navigation-subcategories'` not `:data-testid="'navigation-subcategories' + somePrefix"`, so the regex is not necessary

packages/theme/components/Header/Navigation/__tests__/HeaderNavigationSubcategories.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+// Women category selection
+const childrenProps = { props: { currentCategory: CategoryTreeDataMock[0].children[1] } };
+const noChildrenProps = { props: { currentCategory: CategoryTreeDataMock[0].children[0] } };
+
+describe('HeaderNavigationSubcategories', () => {
+  it('displays level 1 children category links', () => {
+    const { getByRole } = render(HeaderNavigationSubcategories, childrenProps);
+    getByRole('link', { name: /tops/i });
+    getByRole('link', { name: /bottoms/i });
```
sethidden, line 19:
```suggestion
    expect(getByRole('link', { name: /tops/i })).toBeTruthy();
    expect(getByRole('link', { name: /bottoms/i })).toBeTruthy();
```

packages/theme/components/Header/Navigation/__tests__/HeaderNavigation.spec.js
```diff
@@ -0,0 +1,50 @@
+import userEvent from '@testing-library/user-event';
+import HeaderNavigation from '../HeaderNavigation.vue';
+import { render } from '~/test-utils';
+
+import { useUiHelpers } from '~/composables';
+import useUiHelpersMock from '~/test-utils/mocks/useUiHelpersMock';
+import CategoryTreeDataMock from '~/test-utils/mocks/categoryTreeDataMock';
+
+jest.mock('~/composables');
+useUiHelpers.mockReturnValue(useUiHelpersMock());
```
sethidden, line 10: I assume you did this because otheriwse a `route.value` error appears when jest imports useUihelpers

You can initialize route.value like
```ts
const sharedOptions = { props: { categoryTree: CategoryTreeDataMock[0].children }, routes: [{ path: '/' }]}
const initialCall = (vue, router, _store) => {
  router.push('/');
};
```

then in the test

```
  it('displays main categories', () => {
    const { getByRole } = render(HeaderNavigation, sharedOptions, initialCall); // <-- initialCall added
    expect(getByRole('link', { name: /women/i })).toBeTruthy();
    expect(getByRole('link', { name: /gear/i })).toBeTruthy();
    expect(getByRole('link', { name: /training/i })).toBeTruthy();
  });
```

Jest - after the above changes - will render nuxt-links instead of `a role="link"`, which is good because the new categories are also nuxt-links in the app (not `<a href=""` like they are with the current assertions) 
image

sethidden: In the state I've shown above it's impossible to access the nuxt-img, so I guess it's best to add a `data-testid="link"` to HeaderNavigation.vue on the SfLink component

sethidden: The point of all of this is to test as much real code as possible, hence why I'm bothering you with the mocking of useUiState that much

packages/theme/components/Header/Navigation/__tests__/HeaderNavigationSubcategories.spec.js
```diff
@@ -0,0 +1,54 @@
+import userEvent from '@testing-library/user-event';
+import HeaderNavigationSubcategories from '../HeaderNavigationSubcategories.vue';
+import { render } from '~/test-utils';
+import CategoryTreeDataMock from '~/test-utils/mocks/categoryTreeDataMock';
+import { useUiHelpers } from '~/composables';
+import useUiHelpersMock from '~/test-utils/mocks/useUiHelpersMock';
+
+jest.mock('~/composables');
+useUiHelpers.mockReturnValue(useUiHelpersMock());
```
sethidden, line 9:
```suggestion
```

(same as in previous comment)

packages/theme/components/Header/Navigation/HeaderNavigation.vue
```diff
@@ -0,0 +1,77 @@
...
+  },
+  setup() {
+    const { getCatLink } = useUiHelpers();
+    const currentCategory = ref<CategoryTree>(null);
+
+    const setCurrentCategory = (category: CategoryTree | null) => {
+      currentCategory.value = category;
+    };
+
+    const hasChildren = (category: CategoryTree) => Boolean(category?.children);
```
sethidden, line 53:
```suggestion
    const hasChildren = computed(() => Boolean(currentCategory?.children);
```
That way you can avoid calling the function in the template

packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
```diff
@@ -0,0 +1,110 @@
...
+  },
+  props: {
+    currentCategory: {
+      type: Object as PropType<CategoryTree | null>,
+      default: () => null,
+    },
+  },
+  setup() {
+    const { getCatLink } = useUiHelpers();
+    const hasChildren = (category: CategoryTree) => Boolean(category?.children.length > 0);
```
sethidden, line 62: You have the same function in HeaderNavigation.vue, just without the `.length`

.eslintrc.js
bartoszherba: I just did and can, what is the problem? We discussed that on slack.

packages/theme/components/Header/Navigation/__tests__/HeaderNavigation.spec.js
Frodigo: I don't agree with that.  getByRole is implicitly an assertion so it's redundant to wrap it by the expected

packages/theme/components/Header/Navigation/__tests__/HeaderNavigation.spec.js
bartoszherba: No, I want to mock everything but class under test. This is how unit tests work.

bartoszherba: The point of UT is to test ONLY the class under tests, not as much code as possible - i want to test ONLY the code under test in a controlled environment.

packages/theme/components/Header/Navigation/__tests__/HeaderNavigationSubcategories.spec.js
bartoszherba: Nope, mock everything but class under test.

packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
bartoszherba: I do not want create separate file just to have this function.

.eslintrc.js
sethidden: Ah I see in the vue testing library docs that they use it that way aswell, ie getBy* is the same as expect(). Then it'd make sense to register the assertions you use as assert fns with the rule
```
{
  "jest/expect-expect": [
    "error",
    {
      "assertFunctionNames": ["expect", "getByText", ...etc],
      "additionalTestBlockFunctions": []
    }
  ]
}
```

packages/theme/components/Header/Navigation/__tests__/HeaderNavigationSubcategories.spec.js
bartoszherba: Ok, fixed.

packages/theme/components/Header/Navigation/HeaderNavigation.vue
bartoszherba: I think that pure function will be ultimately faster.

Frodigo (approved):

sethidden:
Reported one more issue to Bartek on Slack - GitHub doesn't like .mov files for some reason

sethidden (changes requested):
(see above)

sethidden (dismissed):
(blocking pr for now because I'll be testing this with a screen reader. I'll unblock once I'm done)

packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
```diff
@@ -0,0 +1,228 @@
...
+          @keydown.right.native.prevent="navRight()"
+          @keydown.left.native.prevent="navLeft()"
+          @keydown.down.native.prevent="navDown()"
+          @keydown.up.native.prevent="navUp()"
+        >
+          <SfHeading
+            :level="5"
+            :title="catLvl1.name"
+            class="sf-heading sf-heading--left"
+          />
```
sethidden, line 30:
```suggestion
  <div class="sf-title sf-title--level-5" <!-- the classes aren't actually named like that, just a demo -->
```

Visually setting a heading level 5 makes sense. Semantically though, there's no heading level 4 above this heading level 5, so the heading structure is disorganized.

It's fine to use just a dumb div with a class to keep the visual feel of heading level 5 but leave out the (inaccurate) semantics. It'd be best to have a correct heading structure instead of the div, but the div is fine for now.

VoiceOver reads "heading level 5, link" when hovering this element, but doesn't read heading levels for the top-level categories "Men", "Women" etc.

packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
```diff
@@ -0,0 +1,228 @@
...
+        <SfList v-if="hasChildren(catLvl1)">
+          <SfListItem
+            v-for="(catLvl2, idxLvl2) in catLvl1.children"
+            :key="idxLvl2"
+          >
+            <SfLink
+              ref="lvl2CatRefs"
+              class="header-navigation__link2"
+              :link="localePath(getCatLink(catLvl2))"
+              tabindex="-1"
```
sethidden, line 41: So I understand this disables the tab navigation because there's arrow navigation.
This worries me because:
- on one hand, there's keyboard navigation in the form of keyboard arrows so this should theoretically be fine
- on the other hand, the user will not know that such navigation exists. You can expect the user to know the built-in browser navigation stuff (Tab, Shift-Tab) but downright disabling Tab navigation is a big deal

I'm pretty sure you won't want to rework the entire kbd navigation to support tab, so I'm writing this as a heads up

packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
```diff
@@ -0,0 +1,228 @@
...
+  },
+  props: {
+    currentCategory: {
+      type: Object as PropType<CategoryTree | null>,
+      default: () => null,
+    },
+    hasFocus: {
+      type: Boolean,
+      default: false,
+    },
```
sethidden, line 81:
```suggestion
```
Is there any way to avoid this? The component could render HTML, but the focus information is contained in the browser. 

Something with https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement maybe? ie. if document.activeElement === (this component's ref) then setupNav

bartoszherba: this is a visibility flag

packages/theme/components/Header/Navigation/HeaderNavigation.vue
```diff
@@ -0,0 +1,144 @@
...
+  >
+    <div class="header-navigation__main">
+      <HeaderNavigationItem
+        v-for="(category, index) in categoryTree"
+        :key="index"
+        ref="lvl0CatRefs"
+        :data-testid="category.uid"
+        :label="category.name"
+        :link="localePath(getCatLink(category))"
+        tabindex="0"
```
sethidden, line 14: This applies to the HeaderNavigationItem's root element, which is a div. Divs are not interactive (foucsable with tab) elements by default, unlike the `<a>` (hopefully SfLink's root element) which is interactive by default. Would it be possible to make HeaderNavigationItem have a root element in the form of `<a>` so the tabindex is not needed?

HeaderNavigationSubcategories work fine (focus the link directly)

sethidden: Very ugly but potentially useful - it's possible to access refs of refs, so you could do this.$refs.myRef.$refs.link.focus()

https://stackoverflow.com/a/58358598/10706046

sethidden (commented):
unblocking

Frodigo (approved):
Looks good. for now. Comments reported by @sethidden  will be addressed by other PRs.

packages/theme/components/Header/Navigation/HeaderNavigation.vue
bartoszherba: Thanks but we are using composition API so `this.whatever` will not work.

packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
bartoszherba: Changed that to h2

packages/theme/components/Header/Navigation/HeaderNavigation.vue
bartoszherba: , Root will be now <a> instead of <div>, tabindex not only indicate that something can be tabbed but also tab order so i leave it as it is, just updated the value 0 > 1

packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
sethidden: There's no h1 so h2 doesn't have a parent aswell. This should probably a list. Don't want to make another round of review and since Marcin mentioned he wants this to go in I'll resolve

sethidden (approved):
Accepting based on Marcin's comment about the a11y fixes supposed to be done later

PR #910, opened 2022-04-22
refactor!: components for product list/grid

ghost:
Before this PR, the category page (category.vue)'s product grid and list views were inlined in the category.vue template.

After this PR, both list and grid components are separate. This means the category.vue template is a bit smaller now.

Acceptance criteria

- [x] There is a component responsible for rendering grid mode category page
 - [x] There is a component responsible for rendering list mode category page
 - [x] Components have unit tests

### DIFF
--- a/packages/theme/composables/useCart/useCart.ts
+++ b/packages/theme/composables/useCart/useCart.ts
@@ -42,8 +42,7 @@ type UseCartApplyCouponParams = ComposableFunctionArgs<{
 /**
  * Parameters accepted by the `isInCart` method in the {@link useCart} composable
  */
-type UseCartIsInCartParams<CART, PRODUCT> = {
-  currentCart: CART
+type UseCartIsInCartParams<PRODUCT> = {
   product: PRODUCT
 };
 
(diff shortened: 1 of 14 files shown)


packages/theme/modules/catalog/pages/category.vue
```diff
@@ -31,148 +31,30 @@
           :pagination="pagination"
           @reloadProducts="fetch"
         />
-        <div
-          v-if="$fetchState.pending"
```
sethidden, line 35: todo

packages/theme/modules/catalog/pages/category.vue
```diff
@@ -374,7 +247,58 @@ export default defineComponent({
...
+
+      const reviewProps = {
+        reviewsCount: getTotalReviews(product),
+        scoreRating: getAverageRating(product),
+      };
+
+      const link = context.app.localePath(`/p/${getProductSku(product)}${getSlug(product, product.categories[0])}`);
+
+      /**
+       * Most props of SfProductCard and SfProductCardHorizontal are the same.
```
sethidden, line 281: I think I should move this to the respectable components so they're easier to test.
The downside is more calculations will be done but if you think about it it will be recalculated only when you change product views. And that' s just for 10 products anyway

packages/theme/composables/useCart/useCart.ts
```diff
@@ -42,8 +42,7 @@ type UseCartApplyCouponParams = ComposableFunctionArgs<{
 /**
  * Parameters accepted by the `isInCart` method in the {@link useCart} composable
  */
-type UseCartIsInCartParams<CART, PRODUCT> = {
```
sethidden, line 45: The types did not match the implementation here. The cart argument is no longer necessary

RetroModernDev (approved):

sethidden:
now we wait for the speedcurve results 👀

PR #927, opened 2022-04-26
fix: product wrong swipe icon direction

ghost:
M2-513
image

the icon now bobs side to side instead of up-down

### DIFF
--- a/packages/theme/pages/Product.vue
+++ b/packages/theme/pages/Product.vue
@@ -769,14 +769,15 @@ export default defineComponent({
 }
 
 @keyframes moveicon {
+
   0% {
-    transform: translate3d(0, 0, 0);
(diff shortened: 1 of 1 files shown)


packages/theme/pages/Product.vue
```diff
@@ -769,14 +769,15 @@ export default defineComponent({
 }
 
 @keyframes moveicon {
+
   0% {
-    transform: translate3d(0, 0, 0);
+    transform: translate3d(0, 0, 0) rotate(90deg) scale(1, -1);
```
sethidden, line 774: magically rotating and mirroring it vertically so the "down" icon becomes a "right" icon

RetroModernDev (approved):

PR #930, opened 2022-04-26
feat: write docs and export useApi and its type from ~/composables

VitorLuizC:
## Description

- Allow developers to declare generic params for `query` and `mutate` methods;
- Define a type for `useApi` return;
- Named export `useApi` and `UseApiInterface` from '~/composables';
- Fixed broken docs generator because `useApi` was using DOM types and `context.$graphql`;
- Write TSDoc comments for useApi composable and its type.

## Related Issue

https://vsf.atlassian.net/browse/M2-373

## Types of changes

## Checklist:


sethidden (changes requested):
(blocking because I need to test some alternatives on the local branch)

packages/theme/composables/useApi/UseApi.ts
```diff
@@ -0,0 +1,24 @@
+import type { Variables } from 'graphql-request';
```
sethidden, line 1: I know all the other composables have their own dedicated Use[ComposableName].ts file with the types, but we've actually been speaking about removing those and just merging the `use[ComposableName].ts` code into the `index.ts` file so both the "JS" code and TS types are in a single index.ts file.

The reason for this is this interface isn't reused anywhere - it's not meant to be extended. So it's easier to just make changes in a single index.ts file, rather than in both (if you eg. add some new feature to a composable).

Would it be possible to remove this file and have it all in-one in index.ts? ie. would api-extractor work well with that?

packages/theme/composables/useApi/index.ts
```diff
@@ -1,14 +1,82 @@
 import { useContext } from '@nuxtjs/composition-api';
 import type { Variables } from 'graphql-request';
+import type { UseApiInterface } from './UseApi';
 
-type Request = <T = any, V = Variables>(document: string, variables?: V, requestHeaders?: HeadersInit) => Promise<T>;
```
sethidden, line 4: In the PR you mention
```
Allow developers to declare generic params for query and mutate methods;
```
but this is already possible on `develop`.

Would it be possible to keep it as-is? Ie. `const query : Request = `? The arguments of query and mutate are identical so the type made sense to me when I was writing it

packages/theme/composables/useApi/index.ts
```diff
@@ -1,14 +1,82 @@
...
+ */
+export function useApi(): UseApiInterface {
   const context = useContext();
-  const query : Request = (document, variables, headers) => context.$graphql.query.request(document, variables, headers);
-  const mutate : Request = (document, variables, headers) => context.$graphql.mutation.request(document, variables, headers);
+
+  const query = <DATA = any, VARIABLES extends Variables = Variables>(
+    document: string,
+    variables?: VARIABLES,
+    headers?: HeadersInit,
```
sethidden, line 69: The types are in UseApiInterface so it'd be nice if they didn't need to be duplicated in index.ts. Though with my earlier suggestion of removing UseApi.ts this comment could resolve itself

packages/theme/composables/useApi/index.ts
```diff
@@ -1,14 +1,82 @@
...
+ *     const { query } = useApi();
+ *
+ *     const price = ref<number | null>(null);
+ *     const loading = ref(false);
+ *     const error = ref<Error | null>(null);
+ *
+ *     const getCartPrice = async (cartId: string) => {
+ *       loading.value = false;
+ *       try {
+ *         const result = await query(GET_CART_PRICE_QUERY, { cartId }, {
```
sethidden, line 39:
```suggestion
 *         const result = await query<SomeDummyType>(GET_CART_PRICE_QUERY, { cartId }, {
```
May be worth demonstrating that passing a generic is possible (don't treat the suggestion verbatim, I'm just not sure how to name SomeDummyType :P)

packages/theme/composables/useApi/index.ts
```diff
@@ -1,14 +1,82 @@
...
+ *           'Accept': 'application/json',
+ *         });
+ *
+ *         price.value = result.data?.cart?.prices?.[0]?.subtotal_excluding_tax?.value ?? null;
```
sethidden, line 43: Hmm now that I see the `result.data?` I'm thinking the type of `query` should be something like `<T = any, V = Variables> (document, variables, somearg) => Promise<ApolloResponse<T>>`.

Otherwise users will always have to manually write `query<ApolloResponse<TheActualType>>`

packages/theme/composables/useApi/index.ts
VitorLuizC: Yeap, I clearly misunderstood the generics. I'll rollback this type and use it in the `UseApiInterface`.

packages/theme/composables/useApi/index.ts
VitorLuizC: My example is wrong. I read both [`graphql-request`](https://github.com/prisma-labs/graphql-request) docs and types, and it really resolves the `Promise` with data instead of the response object.

I'll fix my example.

sethidden (approved):

PR #933, opened 2022-04-27
fix: prevent content jump when variant is selected on PDP

bartoszherba:
## Description
prevent content jump when a variant is selected on PDP

#803 
## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/composables/useCart/useCart.ts
+++ b/packages/theme/composables/useCart/useCart.ts
@@ -79,7 +79,7 @@ export interface UseCartInterface<CART, CART_ITEM, PRODUCT> {
    *
    * This is `true` when loading, saving, updating, or removing customer addresses and `false` otherwise.
    */
-  loading: DeepReadonly<Ref<boolean>>;
+  loading: Readonly<Ref<boolean>>;
   /** The error object */
   error: DeepReadonly<Ref<UseCartErrors>>;
 }
--- a/packages/theme/composables/useReview/useReview.ts
+++ b/packages/theme/composables/useReview/useReview.ts
@@ -55,7 +55,7 @@ export interface UseReviewInterface {
   /**
    * Indicates whether any of the methods is in progress
    */
-  loading: DeepReadonly<Ref<boolean>>;
+  loading: Readonly<Ref<boolean>>;
 
   /**
    * Contains errors from any of the composable methods
--- a/packages/theme/pages/Product.vue
+++ b/packages/theme/pages/Product.vue
@@ -1,7 +1,7 @@
 <template>
   <div id="product">
     <SfLoader
-      :class="{ 'loading--product': productDataIsLoading }"
+      class="loading--product"
       :loading="productDataIsLoading"
     >
       <div>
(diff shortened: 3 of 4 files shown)


sethidden (changes requested):
reviewing

packages/theme/composables/useCart/useCart.ts
```diff
@@ -79,7 +79,7 @@ export interface UseCartInterface<CART, CART_ITEM, PRODUCT> {
    *
    * This is `true` when loading, saving, updating, or removing customer addresses and `false` otherwise.
    */
-  loading: DeepReadonly<Ref<boolean>>;
+  loading: Readonly<Ref<boolean>>;
```
sethidden, line 82: With this change:
image


Without this change (previous):
image

Should stay as is imo
```suggestion
  loading: DeepReadonly<Ref<boolean>>;
```

packages/theme/composables/useReview/useReview.ts
```diff
@@ -55,7 +55,7 @@ export interface UseReviewInterface {
   /**
    * Indicates whether any of the methods is in progress
    */
-  loading: DeepReadonly<Ref<boolean>>;
+  loading: Readonly<Ref<boolean>>;
```
sethidden, line 58: Same as in previous comment

packages/theme/pages/Product.vue
```diff
@@ -470,61 +462,76 @@ export default defineComponent({
...
       };
 
-      products.value = await getProductDetails({
+      const result = await getProductDetails({
         ...baseSearchQuery,
       });
 
-      if (product?.value?.length === 0) nuxtError({ statusCode: 404 });
+      product.value = result.items[0] as Product ?? null;
+      if (Boolean(product?.value?.sku) === false) nuxtError({ statusCode: 404 });
```
sethidden, line 485:
```suggestion
      if (!Boolean(product?.value?.sku)) nuxtError({ statusCode: 404 });
```
nitpick, can commit through github

packages/theme/pages/Product.vue
```diff
@@ -470,61 +462,76 @@ export default defineComponent({
...
-      }];
-
-      const categoriesTags = categories.value.map((catId) => ({
-        prefix: CacheTagPrefix.Category,
-        value: catId,
-      }));
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-      addTags([...tags, ...productTags, ...categoriesTags]);
+        const baseReviews = Array.isArray(productReviews)
+          ? [...productReviews].shift()
```
sethidden, line 492:
```suggestion
          ? productReviews[0]
```

I know this is just a diff because you moved the code from above but this is so weird that I felt the need to leave a comment.

shift() removes the first element from the array (mutates the array) and returns the first element. So here a new array is created just for the sake of getting the first element from it. This doesn't even create any new references or anything

packages/theme/pages/Product.vue
```diff
@@ -470,61 +462,76 @@ export default defineComponent({
...
+      const result = await getProductDetails({
         ...baseSearchQuery,
       });
 
-      if (product?.value?.length === 0) nuxtError({ statusCode: 404 });
+      product.value = result.items[0] as Product ?? null;
+      if (Boolean(product?.value?.sku) === false) nuxtError({ statusCode: 404 });
 
-      productReviews.value = await searchReviews(baseSearchQuery);
+      if (firstFetch) {
```
sethidden, line 487: I understand this is here to accomodate the client-side changes (eg. changing color of a clothing item)

If some logic is "firstFetch" it means that it's either client-side or server-side, but if some logic is NOT "firstFetch", it means it's only client-side (ie. you can't change color of clothes on PDP on the serverside - it's always done on the client side).

Imo if something is client-side, it shouldn't be in useFetch, since useFetch is used for bot client-side and server-side logic.

packages/theme/pages/Product.vue
```diff
@@ -470,61 +462,76 @@ export default defineComponent({
...
+          prefix: CacheTagPrefix.Category,
+          value: catId,
+        }));
+
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+        addTags([...tags, ...productTags, ...categoriesTags]);
+      }
     });
 
+    const updateProductConfiguration = (label, value) => {
```
sethidden, line 521:
```suggestion
    const updateProductConfiguration = (label: SomeType, value: SomeType) => {
```

packages/theme/pages/Product.vue
```diff
@@ -136,14 +134,14 @@
                   </SfSelectOption>
                 </SfSelect>
               </template>
-              <template v-if="product.__typename === 'GroupedProduct'">
+              <template v-if="product !== null && product.__typename === 'GroupedProduct'">
```
sethidden, line 137: It's possible to avoid the condition duplication
```suggestion
          <template v-if="product !== null>
              <template v-if="product.__typename === 'GroupedProduct'">
              ...
              </template>
              <template v-else-if="product.__typename === 'someotherproductwhatever'">
              </template>
```

sethidden (changes requested):

packages/theme/pages/Product.vue
sethidden: The more nested templates can be removed aswell by just putting v-if directly on the grouped-product-selector

packages/theme/composables/useCart/useCart.ts
bartoszherba: Screenshot 2022-04-27 at 14 52 25

And Readonly perfectly align with the loading data structure.

/**
 * Make all properties in T readonly
 */
type Readonly<T> = {
    readonly [P in keyof T]: T[P];
};

packages/theme/pages/Product.vue
bartoszherba: ok

packages/theme/pages/Product.vue
bartoszherba: ok

packages/theme/pages/Product.vue
bartoszherba: ok

packages/theme/pages/Product.vue
bartoszherba: Ultimately the whole page must be reworked to align with our modular approach and decomposition of components, but this is why we have tasks to do that so we can do smaller, intermediate steps.

packages/theme/composables/useCart/useCart.ts
bartoszherba: IDK why do you have `const loading: any` in the first place

packages/theme/pages/Product.vue
bartoszherba: I made an improvement to avoid "if"

sethidden (approved):

PR #936, opened 2022-04-28
feat: add skeleton loaders in category navbar

bartoszherba:
## Description
add skeleton loaders in category navbar

## How Has This Been Tested?

- Visit category page, execute any action eg. add filter
- Observe that skeletons are now masking loaded fields

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/components/SkeletonLoader/index.vue
+++ b/packages/theme/components/SkeletonLoader/index.vue
@@ -1,15 +1,13 @@
 <template>
-  <div>
+  <div
+    :class="componentClass"
+    :style="componentStyle"
+  >
     <slot v-if="!isLoading" />
-    <span
-      v-else
-      :class="componentClass"
-      :style="componentStyle"
-    />
   </div>
 </template>
 
-<script>
+<script lang="ts">
 import { computed, defineComponent } from '@nuxtjs/composition-api';
 
 export default defineComponent({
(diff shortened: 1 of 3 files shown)


packages/theme/components/SkeletonLoader/index.vue
```diff
@@ -1,15 +1,13 @@
 <template>
-  <div>
+  <div
```
sethidden, line 2: At this point you can remove componentClass and possibly componentStyle since if you do `<Skeleton loader :style="{ margin: 0 }"` or something the :style and :class will be passed to the root element

This'd be annyoing to replace all skeletons in the whole app but just be aware, no action needed - it's potentially just more flexible since you don't have to add a new prop for margin, width, height

sethidden (changes requested):
CategoryNavbar still has some layout shift on narrow viewports. I'm only talking about the layout shift in CategoryNavbar, not the other ones (sidebar etc), since this PR is only about Category Navbar
https://user-images.githubusercontent.com/5359825/165923395-18646955-6136-4e03-972a-05219edd4e72.mov

packages/theme/components/SkeletonLoader/index.vue
bartoszherba: I really do not want to rework skeleton loaders in this issue ;/
Though I fixed layout issues for a mobile view

Frodigo (approved):

sethidden (approved):

PR #937, opened 2022-04-28
refactor: customer module

sethidden:
move customer related files to a single customer folder


packages/theme/components/LoginModal.vue
```diff
@@ -299,9 +299,9 @@ import {
 import { ValidationProvider, ValidationObserver, extend } from 'vee-validate';
 import { required, email } from 'vee-validate/dist/rules';
 import {
-  useUiState, useWishlist, useUser, useForgotPassword, useCart
+  useUiState, useWishlist, useUser, useForgotPassword, useCart,
 } from '~/composables';
-import { customerPasswordRegExp, invalidPasswordMsg } from '~/helpers/customer/regex';
+import { customerPasswordRegExp, invalidPasswordMsg } from '~/modules/customer/regex/regex';
```
sethidden, line 304: probably change the regex folder name to something like validation-rules

packages/theme/modules/customer/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  this.extendRoutes((routes) => {
+  // {
+  //   name: 'my-account',
+  //   path: '/my-account/:pageName?',
+  //   component: path.resolve(themeDir, 'pages/MyAccount.vue'),
+  // },
+    routes.unshift({
+      name: 'my-account',
+      path: '/my-account/:pageName?',
+      component: path.resolve(path.dirname(url.fileURLToPath(import.meta.url)), 'pages/MyAccount.vue'),
```
sethidden, line 15: This should be lazy loaded, but the task says not to make any functional changees so I don't know

sethidden: Spent a few m trying to get this done and I'd prefer to do this in another task

packages/theme/routes.js
```diff
@@ -11,11 +11,6 @@ export function getRoutes(themeDir = __dirname) {
     path: '/p/:id/:slug/',
     component: path.resolve(themeDir, 'pages/Product.vue'),
   },
-  {
-    name: 'my-account',
```
sethidden, line 15: is now registered by module

packages/theme/components/LoginModal.vue
```diff
@@ -299,9 +299,9 @@ import {
 import { ValidationProvider, ValidationObserver, extend } from 'vee-validate';
 import { required, email } from 'vee-validate/dist/rules';
 import {
-  useUiState, useWishlist, useUser, useForgotPassword, useCart
+  useUiState, useWishlist, useUser, useForgotPassword, useCart,
```
bartoszherba, line 302: Please, import useUser from the module, as we discussed on slac, ~/composables will be reworked only for api-docs purposes. The same for other composables like useForgotPassword etc.
Please, do that for all files that were migrated to the module directory and all files that import scripts that were moved.

bartoszherba:
@sethidden 
I am not able to add a lot of comments to the code that was just moved without any change so I add them here

1. We should add `lang="ts"` wherever is missing in moved files. You will be not able to push changes without it due to linter unless you use --no-verify flag. This will also help you to think about types used and understand which types are closely related to the customer modules and must be moved. I know that we have a separate task to make sure that all components have a proper lang type set but doing it now will instantly increase the quality of the module as well as you will have to close think through changes.

packages/theme/modules/customer/index.ts
```diff
@@ -0,0 +1,20 @@
...
+const nuxtModule : Module = function customerModule() {
+  this.extendRoutes((routes) => {
+    const component = path.resolve(
+      path.dirname(url.fileURLToPath(import.meta.url)),
+      'pages/MyAccount.vue',
+    );
+
+    routes.unshift({
+      name: 'my-account',
+      path: '/my-account/:pageName?',
```
bartoszherba, line 14: I think that we should rename pages, url looks like `<domain>/<store>/my-account/MyProfile` I think that we should keep lowercase and use `-` as a words separator so the url will look like `<domain>/<store>/my-account/my-profile` The same holds for all future modules and routes.

bartoszherba (changes requested):

packages/theme/modules/customer/index.ts
sethidden: I didn't want to touch the routes in this PR because I know that https://vsf.atlassian.net/browse/M2-506 requires me to remove the pageName stuff and create regular routes

There's a mess in myaccount where the pageName is tightly tied to the translation label, so I'll just do it as I fix it in M2-506

sethidden:
> 1. We should add lang="ts" wherever is missing in moved files. You will be not able to push changes without it due to linter unless you use --no-verify flag. This will also help you to think about types used and understand which types are closely related to the customer modules and must be moved. I know that we have a separate task to make sure that all components have a proper lang type set but doing it now will instantly increase the quality of the module as well as you will have to close think through changes.

I'd rather do it as a part of the next task with the MyAccount rework. I'm treating this PR/task more as "move all the files and change the paths". I'm already tired of moving all those files around so I'd prefer to do that as a part of M2-506 where I'll be actually touching the code and can start fresh. I can add it to the acceptance criteria for 506 if you want.

> 2. Make sure that there are no TS errors in any files that were moved/edited

answered in point 1

> DONE 3. Sometimes you import import type { ComposableFunctionArgs } from '~/composables/types'; sometimes from '~/types/core'; please unify the apporach

Replaced the 3 occurences of this in the whole app

> DONE 4. You completely forgot about helpers, eg: import { transformUserCreateAddressInput, transformUserUpdateAddressInput } from '~/helpers/userAddressManipulator'; in packages/theme/modules/customer/composables/useAddresses/index.ts (helper logger and such are exceptions)

> DONE 5. Keep all components in the components directory eg: 
customer/forms -> customer/components/forms

Just keeping in components/ since there's just one group of components (there's no components/forms in the current version of PR)

> DONE 6. Change reference for getter in the same way as for composables, import them from the module not from ~/getters

Did my best to replace all, but might've missed some

> HOLD 7. I am not sure but you said that you have split store and move it into the customer module, I do not see it in the code? import { useCustomerStore } from '~/stores/customer';

I'm not splitting the store - I'm leaving it as is but moving the useUser composable, despite it being shared with useUser, useWishlist and useCart. This is out of scope

> DONE 8. You did not move getters, I think that we must move:
*** DONE userGetters
*** DONE userShippingGetters
*** DONE userBillingGetters
*** DONE userAddressesGetters
*** DONE forgotPasswordGetters
*** DONE maybe addressGetter

sethidden:
I'll still be changing the mocks in the unit tests and fixing the conflicts since Bart's PR went in. But most of the changes are done.

bartoszherba (approved):

PR #939, opened 2022-04-28
refactor: move products related code to the module spaces

bartoszherba:
## Description
The goal is to move all Product-related scripts into the module/catalog/product module

## Types of changes

## Checklist:


packages/theme/components/Header/Navigation/HeaderNavigation.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div class="header-navigation">
+  <div
+    class="header-navigation"
+    @mouseleave="setCurrentCategory(null)"
```
bartoszherba, line 4: This is a bugfix for a menu, I forgot to revert that in the previous PR. This is small so I decided

packages/theme/modules/catalog/index.ts
```diff
@@ -3,11 +3,17 @@ import url from 'node:url';
 import { Module } from '@nuxt/types';
 
 const nuxtModule : Module = function categoryModule() {
+  const themeDir = path.dirname(url.fileURLToPath(import.meta.url));
+
   this.extendRoutes((routes) => {
-    routes.unshift({
+    routes.push({
```
sethidden, line 9: The unshift kinda makes sense here - append the route to the beginning so that it overrides whatever is in the routes.js comes later, so the module route has the highest priority

packages/theme/modules/catalog/product/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  GroupedProduct as BaseGroupedProduct, ProductAttributeFilterInput, ProductAttributeSortInput,
+  ProductInterface,
+  VirtualProduct as BaseVirtualProduct,
+} from '~/modules/GraphQL/types';
+
+/**
+ * There is no __typename in GraphQL definitions but type_id is marked as a deprecated
+ */
+export interface Product extends ProductInterface, ConfigurableProduct, Omit<BundleProduct, 'items'>, Omit<GroupedProduct, 'items'>, Omit<DownloadableProduct, 'items'>, Omit<VirtualProduct, 'items'> { __typename: string }
+export interface BundleProduct extends BaseBundleProduct { __typename: string }
```
sethidden, line 14:
```suggestion
export interface BundleProduct extends BaseBundleProduct & { __typename: string }
```

huh kinda surprised that this works without the `&`, if it *really* works then it's fine

packages/theme/modules/catalog/product/types.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface ConfigurableProduct extends BaseConfigurableProduct { __typename: string }
+export interface DownloadableProduct extends BaseDownloadableProduct { __typename: string }
+export interface ConfigurableProduct extends BaseConfigurableProduct { __typename: string }
+export interface GroupedProduct extends BaseGroupedProduct { __typename: string }
+export interface VirtualProduct extends BaseVirtualProduct { __typename: string }
```
sethidden, line 19:
```suggestion
export interface VirtualProduct extends BaseVirtualProduct { __typename: string }
```

You could avoid creating all those interfaces by just creating a generic type
```ts
type WithTypename<TProduct> = TProduct & { __typename: string };
```
and then in code using them like
```
product: Object as PropType<WithTypename<BaseConfigurableProduct>>
```

If there is some "super base product interface shared between all products" (knowing graphql codegen probably not), you could also do `type WithTypename<TProduct extends SuperBaseProduct>` aswell

sethidden (commented):
Left some quality of life comments. Let me know what you think. Once we resolve it's ok to merge

packages/theme/modules/catalog/index.ts
bartoszherba: Fair enough

packages/theme/modules/catalog/product/types.ts
bartoszherba: Outdated.

sethidden (approved):

PR #940, opened 2022-04-28
refactor: use localeroute instead of string path

ghost:
If someone decides to change the path's URL segment, the links will break if they're constructed as a string. However the "name" is path-agnostic

Plus, it's a little easier on the eyes than reaching the crusty template interpolation

There's no task for this, but I didn't want to start working on refactoring the my account module till the customer module is merged.

### DIFF
--- a/packages/theme/helpers/cart/addToCart.ts
+++ b/packages/theme/helpers/cart/addToCart.ts
@@ -25,8 +25,18 @@ export const useAddToCart = () => {
         break;
       case 'BundleProduct':
       case 'ConfigurableProduct':
-        const path = `/p/${productGetters.getProductSku(product)}${productGetters.getSlug(product, product.categories[0])}`;
-        await router.push(String(app.localePath(path)));
+        const sku = productGetters.getProductSku(product);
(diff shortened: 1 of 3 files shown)


packages/theme/helpers/cart/addToCart.ts
```diff
@@ -25,8 +25,18 @@ export const useAddToCart = () => {
         break;
       case 'BundleProduct':
       case 'ConfigurableProduct':
-        const path = `/p/${productGetters.getProductSku(product)}${productGetters.getSlug(product, product.categories[0])}`;
-        await router.push(String(app.localePath(path)));
+        const id = productGetters.getProductSku(product);
```
bartoszherba, line 28: SKU is not an id, there is an id on the product, there is a problem with naming, in route it is named as an ID but we use SKU, that id is probably because of the legacy agnostic code. Not sure what to do here. Maybe at least update const id -> const sku?

bartoszherba (dismissed):
Non critical comment

bartoszherba (approved):

PR #941, opened 2022-04-28
feat: replaced images from cdn.shopify by images from cloudinary

AlexanderDevitsky:
## Related Issue

https://vsf.atlassian.net/browse/M2-408

## Types of changes

## Checklist:


packages/theme/components/CallToAction.vue
```diff
@@ -0,0 +1,43 @@
+<script>
```
bartoszherba, line 1: Please add `lang="ts`
Do not push changes with the `--no-verify` flag.

packages/theme/components/CallToAction.vue
```diff
@@ -0,0 +1,43 @@
...
+export default defineComponent({
+  name: 'CallToAction',
+  extends: ExtendedCallToAction,
+  props: {
+    imageTag: {
+      type: String,
+      default: null,
+    },
+    nuxtImgConfig: {
+      type: Object,
```
bartoszherba, line 19: Please add object type `PropType<type>`

packages/theme/components/CallToAction.vue
```diff
@@ -0,0 +1,43 @@
...
+  computed: undefined,
+};
+
+export default defineComponent({
+  name: 'CallToAction',
+  extends: ExtendedCallToAction,
+  props: {
+    imageTag: {
+      type: String,
+      default: null,
```
bartoszherba, line 16:
```suggestion
      default: '',
```

packages/theme/components/Hero.vue
```diff
@@ -46,27 +46,41 @@ export default defineComponent({
...
+    },
+    image: {
+      type: [Object, String],
       default: '',
     },
-    desktopImage: {
+    imageTag: {
       type: String,
-      default: '',
+      default: null,
```
bartoszherba, line 57:
```suggestion
      default: '',
```
('' === null) === false && (typeof null === 'string') === false

packages/theme/components/Hero.vue
```diff
@@ -46,27 +46,41 @@ export default defineComponent({
...
     },
     link: {
       type: String,
       default: null,
     },
-    mobileImage: {
-      type: String,
-      default: '',
+    nuxtImgConfig: {
+      type: Object,
```
bartoszherba, line 64: Please add PropType

packages/theme/components/Hero.vue
```diff
@@ -46,27 +46,41 @@ export default defineComponent({
...
     },
   },
   setup(props) {
-    const styles = ref({
-      '--hero-background-image': `url('${props.desktopImage}')`,
-      '--hero-background-image-mobile': `url('${props.mobileImage}')`,
-      'background-color': props.background,
-    });
+    const { app } = useContext();
+    const image = props.image;
```
bartoszherba, line 70: Can you use SfImage instead, so we avoid custom implementation? SfImage has everything you need and will align with all other parts of application. SfImage supports nuxt image.

packages/theme/pages/Home.vue
```diff
@@ -7,8 +7,11 @@
       :button-text="$t('Learn more')"
       link="/c/women.html"
       background="#eceff1"
-      desktop-image="https://cdn.shopify.com/s/files/1/0407/1902/4288/files/bannerB_1240x400.jpg"
-      mobile-image="https://cdn.shopify.com/s/files/1/0407/1902/4288/files/bannerB_328x224.jpg"
+      image-tag="nuxt-img"
+      :image="{
+        desktop: '/w_1240,h_400,c_fit/homepage/bannerB.webp',
```
bartoszherba, line 12: This is hard to use, please use SfImage as suggested above.

packages/theme/pages/Home.vue
```diff
@@ -39,18 +43,19 @@
         link="/c/women.html"
       />
     </LoadWhenVisible>
+    <CallToAction
+      :title="$t('Subscribe to Newsletters')"
+      :button-text="$t('Subscribe')"
+      :description="$t('Be aware of upcoming sales and events. Receive gifts and special offers!')"
+      image-tag="nuxt-img"
+      :image="{
```
bartoszherba, line 51: Please use SfImage instead of a custom implementation.

packages/theme/pages/Home.vue
```diff
@@ -39,18 +43,19 @@
...
+        desktop: '/w_1240,h_202,c_fit/homepage/newsletter.webp',
+        mobile: '/w_400,h_200,c_fit/homepage/newsletter.webp'
+      }"
+      class="call-to-action"
+    />
     <LoadWhenVisible>
```
bartoszherba, line 57: Please move CTA back inside the LWV component, or if it yields issues (why?) then remove empty LWV component.

bartoszherba (changes requested):
@AlexanderDevitsky 
I think that the main issue with this PR is that you implemented your own nuxt-img solution but we already have SfImage which covers all needs. You can find an example here `packages/theme/pages/Checkout/Payment.vue` 
```<SfImage
            image-tag="nuxt-img"
            :src="getMagentoImage(cartGetters.getItemImage(product))"
            :alt="cartGetters.getItemName(product)"
            :width="imageSizes.checkout.imageWidth"
            :height="imageSizes.checkout.imageHeight"
            :nuxt-img-config="{
              fit: 'cover',
            }"
          />```

AlexanderDevitsky:
@bartoszherba Thank you for the review! I have a few more questions. What if we need to load images of different sizes for desktop and mobile devices. In this case, we need to use two `<SfImage>` components that will have different `width` and `height` props for desktop and mobile? Also this implementation has been done to use `background-image`. Should I replace logic and use `SfImage` component with `object-fit: cover` property for the image in this case?

bartoszherba:
@AlexanderDevitsky This is a hard question, we are using CDN or Redis or both, in such case the pattern will be to determine if we are on mobile or desktop in `mounted` hook and with some ifology display only the mobile or desktop content. If we hide stuff with CSS we will over fetch images in two sizes. I am not sure if this is in the scope of this task anyway, we have a lot of such places that are not covered yet. On mounted hook will be able to return proper device because it will be executed once all cache/cdn html/scripts are already loaded.

And yes please, replace all images with SfImage, configure it as a nuxt image. Should you use cover? IDK, it depends. Eg. on PDP we do not crop images but I imagine that in a banner, the image must take defined space and some should be cropped. So It is up to the context. In most we use `cover` though but maybe `inside` will be more appropriate in a banner? IMO we can use fit, it is a configurable thing and the final implementation can be very easily adjusted.

AlexanderDevitsky:
@bartoszherba I have added some changes. Please check if this is enough

bartoszherba (approved):

PR #950, opened 2022-05-05
refactor: move checkout related scripts to the module space

bartoszherba:
## Description
- create checkout module
- move composables to the checkout module
- move getters to the checkout module

## Types of changes

## Checklist:


packages/theme/modules/checkout/index.ts
```diff
@@ -0,0 +1,46 @@
+import { Module } from '@nuxt/types';
+import path from 'node:path';
+import url from 'node:url';
+
+const nuxtModule : Module = function checkoutModule() {
+  const themeDir = path.dirname(url.fileURLToPath(import.meta.url));
```
sethidden, line 6:
```suggestion
  const moduleDir = path.dirname(url.fileURLToPath(import.meta.url));
```
nitpick: I know it's like that in the other modules but this is not the theme directory. Since you're appending `pages/Checkout.vue`, this variable points to `theme/modules/checkout` (so not theme dir, but the module dir)

sethidden (approved):

packages/theme/modules/checkout/index.ts
bartoszherba: Ok, i fixed that.

sethidden (approved):

PR #953, opened 2022-05-06
refactor: category page - rework useFacet

bartoszherba:
## Description
- move breadcrumbs logic to the breadcrumbs component
- rework useFacet composable

## Types of changes

## Checklist:


packages/theme/modules/catalog/category/components/cms/categoryContent.gql.ts
```diff
@@ -1,9 +1,9 @@
 import { gql } from 'graphql-request';
 
 export default gql`
-  query category($id: Int!) {
-    category(id: $id) {
-      id
+  query getCategoryContentData($filters: CategoryFilterInput) {
+    categoryList(filters: $filters) {
```
bartoszherba, line 5: category is deprecated, i fixed that here.

packages/theme/modules/catalog/category/components/views/useProductsWithCommonCardProps.ts
```diff
@@ -47,7 +47,7 @@ export const useProductsWithCommonProductCardProps = (products: Ref<Product[]>)
       scoreRating: getAverageRating(product),
     };
 
-    const link = context.app.localePath(`/p/${getProductSku(product)}${getSlug(product, product.categories[0])}`);
```
sethidden, line 50: Do you mind using the less string-operation intensive object-based approach here?
```ts
context.app.localeRoute({ name: 'product', params: { id: getProductSku(product), slug: getSlug(product, product.categories[0])
```
You may need to slice the getSlug(product, product.categories[0]) to get the redundant '/' from the beginning

packages/theme/modules/catalog/category/composables/useFacet/SortingOptions.ts
```diff
@@ -0,0 +1,35 @@
+export enum SortingOptionsValuesEnum {
+  DEFAULT = '',
+  NAME_ASC = 'name_ASC',
+  NAME_DESC = 'name_DESC',
+  PRICE_ASC = 'price_ASC',
+  PRICE_DESC = 'price_DESC',
+}
```
sethidden, line 7: optiona: Given the names of the keys largely correspond with the values you could use a string union instead of an enum

```ts
export type SortingOptionsValuesEnum = "" | "name_ASC" | "name_DESC" | "price_ASC" | "price_DESC"
```

(you lose the description of "" as "DEFAULT" though). I wrote this because there's a lot of critique of TS enums, ie. they get compiled down to objects so they contribute to the bundle size, where string unions are just JS strings.

packages/theme/modules/catalog/category/composables/useFacet/index.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * The `useFacet()` composable allows searching for products using facets.
+ *
+ * What makes it powerful is the ability to accept multiple filters, allowing to
+ * narrow down the results to a specific category, search term, etc.
+ */
+export function useFacet(): UseFacetInterface {
+  const { query } = useApi();
+  const loading = ref(false);
+  const result = ref<UseFacetSearchResult<any>>({ data: null, input: null });
```
sethidden, line 24: Can a better type be used here? I get that you don't know what type will come back, but consider:

1. Adding a generic type argument to the useFacet composition function  
```ts
export function useFacet<TData, TInput>() {
   ref<UseFacetSearchResult<data: TData, input: TInput>
}
```
or just `<TResult extends { data: unknown | null, input: unknown | null }>` (note the unknown isntead of any meaning "I don't know what types are there - force me to do some type checks like `if(typeof x === string)` before attempting any operation instead of the any behavior of "just let me do anything")

packages/theme/modules/catalog/category/composables/useFacet/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      const { products } = await query(GetFacetDataQuery, productSearchParams);
+
+      Logger.debug('[Result]:', { products });
+
+      result.value.data = {
+        items: products?.items ?? [],
+        total: products?.total_count,
+        availableSortingOptions: SortingOptions,
+        perPageOptions: PerPageOptions,
```
sethidden, line 54: not sure if this single usage of PerPageOptions warrants a separate file. keeping it as a non-exported const seems fine

packages/theme/modules/catalog/category/getters/facetGetters.ts
```diff
@@ -0,0 +1,64 @@
...
+      ? 1
+      : Math.ceil(searchData.data.total / searchData.input.itemsPerPage)
+  ) : 1;
+
+  return {
+    currentPage: (searchData?.input?.page > totalPages ? 1 : searchData?.input?.page) || 1,
+    totalPages,
+    totalItems: (searchData?.data?.total) ? searchData.data.total : 0,
+    itemsPerPage: searchData?.input?.itemsPerPage || 10,
+    pageOptions: [10, 50, 100],
```
sethidden, line 54: since there's a dedicated page options (if it's deemed still necessary) with the [10,20,30] array, this could be moved to that file?

packages/theme/modules/catalog/pages/category.vue
```diff
@@ -217,10 +212,6 @@ export default defineComponent({
 
       // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
       addTags([...tags, ...productTags]);
-
-      if (!isCategoryTreeLoaded.value) {
-        await loadCategoryTree();
-      }
```
sethidden, line 223: I'm worried that traverseCategoryTree won't be be able to access the categories because they won't be loaded.

I know the sidebar also loads categories in the catalog, but the loading inside category.vue (page root) makes the most sense since it's the entrypoint.

It seems fragile to rely on the categories/filters to load the categories for category.vue.

This should be probably done in router middleware anyway but I'd keep this for now

packages/theme/modules/catalog/category/components/breadcrumbs/CategoryBreadcrumbs.vue
```diff
@@ -1,35 +1,53 @@
...
+      }));
+    });
 
     return {
       breadcrumbs,
     };
   },
 });
 </script>
+<style lang="scss">
```
sethidden, line 45:
```suggestion
<style lang="scss" scoped>
```

sethidden (changes requested):

packages/theme/modules/catalog/category/composables/useFacet/index.ts
sethidden: In a later comment I found another instance of perPageOptions and suggested to maybe move it to that file, so there are 2 options: keep 2 instances in the file, or remove the file altoghether and inline the code

packages/theme/modules/catalog/category/composables/useFacet/SortingOptions.ts
bartoszherba: I know about the issues with enums and I am aware of what you propose but that would be the third convention for enums in this project so let's keep it as it is and eventually rework all places in the future.

packages/theme/modules/catalog/category/composables/useFacet/index.ts
bartoszherba: From the customizability standpoint, it is better to have config things in a separate file. We need to think a little wider, like - can we expect to reuse this in the project? Definitely, we can and that's it. It can be even used somewhere else for calculations or other stuff. Keep responsibilities separated is a good approach.

packages/theme/modules/catalog/category/getters/facetGetters.ts
bartoszherba: I used PerPageOptoions for the consistency.

packages/theme/modules/catalog/pages/category.vue
bartoszherba: It does not have to be loaded, the active category is a computed prop based on the router path. An active category is the only thing I need on the category page at this point.

packages/theme/modules/catalog/category/components/breadcrumbs/CategoryBreadcrumbs.vue
bartoszherba: oks, fixed

packages/theme/modules/catalog/category/components/views/useProductsWithCommonCardProps.ts
bartoszherba: No, I don't mind, it's your code ;)))

sethidden: It's not mine :P I'm just the last in the blame

sethidden (approved):

sethidden (approved):

PR #958, opened 2022-05-09
fix: coupon code doesn't show error

ghost:
1. Add error state for couponcode - ie. when you input a wrong coupon code an error message is shown
2. Refactor coupon code logic to be simpler
3. Add some margins for prettier spacing
4. Fix "promoCodeIsApplied passed to 'disabled' prop should be boolean, but string passed" console error

### DIFF
--- a/packages/theme/components/CouponCode.vue
+++ b/packages/theme/components/CouponCode.vue
@@ -1,41 +1,41 @@
 <template>
   <div>
-    <span
-      v-if="promoIsApplied"
-      class="applied-coupon"
-    >
+    <span v-if="isCouponCodeApplied">
       {{ $t('Applied Coupon') }}:
-      <span class="applied-coupon__code">{{ promoCode }}</span>
+      <span
+        class="applied-coupon-code"
+        v-text="couponCodeAppliedToCart"
+      />
     </span>
-    <div class="promo-code">
+    <div class="coupon-code">
       <SfInput
-        v-model="promoCode"
-        name="promoCode"
-        :disabled="promoIsApplied"
+        v-model="couponCodeUserInput"
+        name="couponCode"
+        :disabled="isCouponCodeApplied"
         :label="$t('Enter promo code')"
-        class="sf-input--filled promo-code__input"
+        class="sf-input--filled coupon-code__input"
+        :error-message="$t('An error occurred')"
+        :valid="!hasAnyError"
       />
       <SfButton
-        class="promo-code__button"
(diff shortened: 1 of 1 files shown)


packages/theme/components/CouponCode.vue
```diff
@@ -44,29 +44,36 @@ export default defineComponent({
...
+    const verifiedPromoCodeAppliedToCart = computed(() => cartGetters.getAppliedCoupon(cart.value)?.code);
+    const isPromoCodeApplied = computed(() => verifiedPromoCodeAppliedToCart.value !== undefined);
 
-    const handleCoupon = async () => {
-      await (promoIsApplied.value
-        ? removeCoupon({ currentCart: cart.value })
-        : applyCoupon({ couponCode: promoCode.value }));
+    const applyOrRemovePromoCode = async () => {
+      try {
+        // eslint-disable-next-line unicorn/prefer-ternary
```
bartoszherba, line 58: Why not follow this rule?

packages/theme/components/CouponCode.vue
```diff
@@ -44,29 +44,36 @@ export default defineComponent({
...
+    const {
+      cart, applyCoupon, removeCoupon, error,
+    } = useCart();
 
-    const setCartCoupon = () => {
-      promoCode.value = promoIsApplied.value;
-    };
+    const promoCodeUserInput = ref('');
+
+    const verifiedPromoCodeAppliedToCart = computed(() => cartGetters.getAppliedCoupon(cart.value)?.code);
```
bartoszherba, line 53: `verifiedPromoCodeAppliedToCart` isn't this name hmm.. too overwhelming? Maybe promoCodeAppliedToCart or even `appliedCoupon` ? If the coupon is applied then it is verified so applied === verified; there is no way to apply the coupon to anything but cart so basically appliedCoupon === couponAppiliedToCart

bartoszherba (dismissed):

packages/theme/components/CouponCode.vue
sethidden: was on the fence on this one because it looks weird as it awaits the ternary result (this is a valid op, but is diffucult to read)

I extracted the ternary to a variable than awaited it

packages/theme/components/CouponCode.vue
sethidden: I additionally changed the convention from promoCode to

bartoszherba (approved):

PR #982, opened 2022-05-12
perf: make read reviews loaded on request

bartoszherba:
## Description
- change loading revies behavior, now will be loaded on explicit action

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/modules/catalog/pages/product.vue
+++ b/packages/theme/modules/catalog/pages/product.vue
@@ -67,7 +67,7 @@
                 </div>
                 <SfButton
                   class="sf-button--text"
-                  @click="changeTab(2)"
+                  @click="changeTab(tabConfig.reviews)"
                 >
                   {{ $t('Read all reviews') }}
                 </SfButton>
@@ -368,6 +368,11 @@ export default defineComponent({
     const { getProductDetails, loading: productLoading } = useProduct();
     const { addItem, loading } = useCart();
 
+    const tabConfig = {
+      description: 1,
+      reviews: 2,
+      additional_info: 3,
+    };
     const {
       search: searchReviews,
       loading: reviewsLoading,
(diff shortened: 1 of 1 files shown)


packages/theme/modules/catalog/pages/product.vue
```diff
@@ -452,17 +457,50 @@ export default defineComponent({
       }
     });
 
+    const { params: { id } } = route.value;
+    const baseSearchQuery = () => ({
```
sethidden, line 461:
```suggestion
    const getBaseSearchQuery = () => ({
```

This is a function, not a variable (I know you just moved the code)

packages/theme/modules/catalog/pages/product.vue
```diff
@@ -452,17 +457,50 @@ export default defineComponent({
...
+      ) as string[],
+    });
+
+    let lastReviewsQuery = '';
+    const fetchReviews = async () => {
+      const newQuery = baseSearchQuery();
+      const stringNewQuery = JSON.stringify(newQuery);
+      if (lastReviewsQuery === stringNewQuery) return;
+
+      lastReviewsQuery = stringNewQuery;
```
sethidden, line 478: I think the manual checking if the new query changed could be avoided. I know you want to avoid redownloading the same thing twice, but I was wondering if it'd be possible to:
1. move the lastReviewsQuery == stringNewQuery logic from fetchReviews, and keep only the regular "just fetch the reviews" behavior inside
2. the lastReviewsQuery is then put into a vue `watch` like:
```ts
watch('$route.query', (newValue, oldValue) =>
   JSON.stringify(newValue) === JSON.stringify(oldValue) || fetchReviews() ) // using the || trick because i want to write this quickly in CR :P
// or turn baseSearchQuery into a computed and watch the computeg baseSearchQuery, since it relies on query aswell

sethidden (changes requested):

packages/theme/modules/catalog/pages/product.vue
sethidden: In the next comment I also mention the possibility of using a computed - in that case, this comment doesn't apply obviously

bartoszherba: Fixed

packages/theme/modules/catalog/pages/product.vue
bartoszherba: As we discussed on slack, fixed

sethidden (approved):
lgtm

PR #996, opened 2022-05-15
feat: show configurable option values in order history AND feat: show more order details

ghost:
This is M2-510 and M2-509 in one PR because I didn't want to create conflicts with myself.

### DIFF
--- a/packages/theme/modules/customer/pages/MyAccount/OrderHistory/OrderSummaryAddressRow.vue
+++ b/packages/theme/modules/customer/pages/MyAccount/OrderHistory/OrderSummaryAddressRow.vue
@@ -1,0 +1,58 @@
+<template>
+  <OrderSummaryRow>
+    <template #label>
+      <slot name="label" />
+    </template>
+    <template #value>
+      <div>
+        {{ address.firstname }} {{ address.lastname }}
+      </div>
+      <div>
+        {{ address.street.join(' ') }}
+      </div>
+      <div>
+        {{ address.city }},
+        {{ address.region }},
+        {{ address.postcode }},
+      </div>
+      <div>
+        {{ countryName }},
+      </div>
+      <div>
+        {{ address.telephone }}
+      </div>
+    </template>
+  </OrderSummaryRow>
+</template>
+
+<script lang="ts">
+import { defineComponent, computed, PropType } from '@nuxtjs/composition-api';
+import type { CustomerOrder, Country } from '~/modules/GraphQL/types';
+import OrderSummaryRow from '~/modules/customer/pages/MyAccount/OrderHistory/OrderSummaryRow.vue';
+
+type Address = CustomerOrder['billing_address'] | CustomerOrder['shipping_address'];
+export default defineComponent({
+  components: {
+    OrderSummaryRow,
+  },
+  props: {
+    address: {
+      type: Object as PropType<Address>,
+      required: true,
+    },
+    countries: {
+      type: Array as PropType<Country[]>,
+      required: true,
+    },
+  },
+  setup(props) {
+    const countryName = computed(
+      () => props.countries.find(({ id }) => props.address.country_code === id).full_name_locale,
+    );
+
+    return {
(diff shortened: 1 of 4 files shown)


packages/theme/modules/customer/pages/MyAccount/OrderHistory/OrderSummaryAddressRow.vue
```diff
@@ -0,0 +1,46 @@
...
+      </div>
+      <div>
+        {{ address.street.join(' ') }}
+      </div>
+      <div>
+        {{ address.city }},
+        {{ address.region }},
+        {{ address.postcode }},
+      </div>
+      <!-- TODO Country  code should be the actual country name -->
```
sethidden, line 18: todo

packages/theme/modules/customer/pages/MyAccount/OrderHistory/SingleOrder.vue
```diff
@@ -24,24 +24,88 @@
             <SfTableHeader>{{ $t('Price') }}</SfTableHeader>
           </SfTableHeading>
           <SfTableRow
-            v-for="item in currentOrder.items"
-            :key="item.key"
```
sethidden, line 28: There's no longer any formatting in the computed as in the last PR - I'm just using values from the Order directly

packages/theme/modules/customer/pages/MyAccount/OrderHistory/SingleOrder.vue
```diff
@@ -24,24 +24,88 @@
...
+            v-for="item in order.items"
+            :key="item.product_sku"
           >
             <SfTableData class="products__name">
-              {{ item.name }}
+              {{ item.product_name }}
```
sethidden, line 31: This is for M2-510

packages/theme/modules/customer/pages/MyAccount/OrderHistory/SingleOrder.vue
```diff
@@ -24,24 +24,88 @@
...
-            class="order-summary"
-          >
-            <SfTableData class="order-summary__spacer" />
-            <SfTableData>{{ property.name }}</SfTableData>
-            <SfTableData>{{ property.value }}</SfTableData>
+            <SfTableData>{{ item.quantity_ordered }}</SfTableData>
+            <SfTableData>{{ $fc(item.product_sale_price.value) }}</SfTableData>
           </SfTableRow>
+
+          <OrderSummaryRow>
```
sethidden, line 44: This is no very large because there's no v-for as before. This is because I needed to add the shipping and billing address components, so the data format is not homogenous anymore

I could probably still do something like

```
v-if="isADdress"
v-else
```

but at this point I really prefer the flexibility of keeping it that way in the template.

I wanted to move it to a separate component but this isn't possible since Vue 2 doesn't support multi-root components (natively). This could be made possible with some more digging through the summary but I didn't want to do this in this PR.

sethidden: Ah, but I forgot could also use `<component :i=""`

packages/theme/modules/customer/pages/MyAccount/OrderHistory/SingleOrder.vue
```diff
@@ -68,57 +141,36 @@ export default defineComponent({
...
+      const orderResult = await search({ filter: { number: { eq: props.orderId } } });
+      const order = orderResult.items[0] ?? null;
 
-    const currentOrder = computed(() => {
-      const order = currentOrderRawData.value?.items[0] ?? null;
-
-      if (order === null) {
-        return null;
-      }
+      const uniqueCountryCodePromises = [...new Set([order.shipping_address.country_code, order.billing_address.country_code])]
```
sethidden, line 156: no way of downloading only selected country codes, it's either load() all countries, or search() one country at a time

bartoszherba (approved):

PR #999, opened 2022-05-16
feat: add skeleton loaders for address edit and addresses details pages

ghost:
M2-576

Show SkeletonLoaders when AddressEdit and AddressesDetails's useFetch is running. This PR doesn't fix the issue where when the route component is lazy-loading a white screen is displayed - I don't think this is fixable since Nuxt 2 doesn't have a Suspense component, right?

### DIFF
--- a/packages/theme/modules/customer/pages/MyAccount/AddressesDetails/AddressForm.vue
+++ b/packages/theme/modules/customer/pages/MyAccount/AddressesDetails/AddressForm.vue
@@ -19,6 +19,7 @@
             required
             :valid="!errors[0]"
             :error-message="$t(errors[0])"
+            :disabled="loading"
           />
         </ValidationProvider>
         <ValidationProvider
@@ -34,6 +35,7 @@
             required
             :valid="!errors[0]"
             :error-message="$t(errors[0])"
+            :disabled="loading"
           />
         </ValidationProvider>
       </div>
@@ -50,6 +52,7 @@
           required
           :valid="!errors[0]"
           :error-message="$t(errors[0])"
+          :disabled="loading"
         />
       </ValidationProvider>
       <SfInput
@@ -72,6 +75,7 @@
             required
             :valid="!errors[0]"
             :error-message="$t(errors[0])"
+            :disabled="loading"
           />
         </ValidationProvider>
         <ValidationProvider
@@ -85,10 +89,10 @@
             v-model="form.region.region"
             v-e2e="'shipping-state'"
             :label="$t('State/Province')"
-            :disabled="!form.country_code"
             name="state"
             :valid="!!form.country_code"
             :error-message="!form.country_code ? $t('Please select a country first') : ''"
+            :disabled="!form.country_code || loading"
           />
           <SfSelect
             v-else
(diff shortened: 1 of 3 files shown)


packages/theme/modules/customer/pages/MyAccount/AddressesDetails/AddressesDetails.vue
```diff
@@ -6,7 +6,12 @@
           v-t="'Manage addresses'"
           class="message"
         />
+        <SkeletonLoader
```
bartoszherba, line 9: Dear @sethidden ;)
Skeleton should be a temporary replacement for a part of the layout in the shape of this layout including blocks and its subparts. One huge skeleton is no better than any other loader. Please, rework it so it will look more like the actual blocks, let say that we display 3x repeated pattern.

packages/theme/modules/customer/pages/MyAccount/AddressesDetails/AddressesDetails.vue
```diff
@@ -6,7 +6,32 @@
           v-t="'Manage addresses'"
           class="message"
         />
+        <template v-if="$fetchState.pending">
```
bartoszherba, line 9: These skeletons does not reflect the actual view. If it does not reflect the layout then it is nothing more but a simple loader, but as for a loader it looks bad :) Please, rework skeletons layout to match the actual layout more closely (does not have to be perfect, just to reflect the layout)

bartoszherba (changes requested):

packages/theme/modules/customer/pages/MyAccount/AddressesDetails/AddressForm.vue
```diff
@@ -200,17 +211,17 @@ import { required, min, oneOf } from 'vee-validate/dist/rules';
...
   onBeforeMount,
 } from '@nuxtjs/composition-api';
 import omitDeep from 'omit-deep';
 import { PropType } from 'vue';
-import { CustomerAddress } from '~/modules/GraphQL/types';
-import { useCountrySearch } from '~/composables';
+import { CountryCodeEnum, useCountrySearch } from '~/composables';
 import type { Countries, Country, UseCountrySearchParams } from '~/composables';
 import addressGetter from '~/modules/customer/getters/addressGetter';
+import { TransformedCustomerAddress } from '~/modules/customer/composables/types';
```
bartoszherba, line 224:
```suggestion
import type { TransformedCustomerAddress } from '~/modules/customer/composables/types';
```

bartoszherba (changes requested):

packages/theme/modules/customer/pages/MyAccount/AddressesDetails/AddressesDetails.vue
sethidden: image

image

Can you be more specific?

sethidden: The default shipping and billing address labels are not reflected in the address because I can't know if an address will be a both a shipping and billing address, or just one of those, or none.

The country code (last skeleton) is longer because currently the address displays the country code, but I think this is wrong, as it'll probably show the country name in the future.

sethidden: image

(the rightmost skeleton is not visible on mobile because of the "x")

bartoszherba (approved):

PR #1000, opened 2022-05-16
refactor!: use order.number instead of deprecated order.order_number

ghost:
M2-581

I didn't want to fix that in my earlier M2-509 PRs because I saw order_number is also used in the Payment.vue components, but it turned out upon closer inspection that order_number is not deprecated when making an order, it's only deprecated in CustomerOrder

Marked as breaking change since this affects api-client and probably needs a release of api-client

We weren't supposed to touch api-client much so I gues I'll just ping @Frodigo and wait :P

### DIFF
--- a/packages/api-client/src/api/customerOrders/customerOrders.ts
+++ b/packages/api-client/src/api/customerOrders/customerOrders.ts
@@ -5,7 +5,6 @@ export default gql`
     customer {
       orders(currentPage: $currentPage, filter: $filter, pageSize: $pageSize) {
         items {
-          order_number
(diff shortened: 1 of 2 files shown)


bartoszherba (approved):

bartoszherba:
@sethidden 
"We weren't supposed to touch api-client much so I guess I'll just ping @Frodigo and wait :P"
And why is it that, we have even tasks to improve queries? And removing deprecated things, especially achieved as easily as this one, is a way to go. Please, finalize this PR.

PR #1008, opened 2022-05-16
chore: fix ~50 eslint errors

sethidden:
Second batch of manual eslint fixes. Error/warning count is 140 -> 96. These ones had to be made by hand.


.eslintrc.js
```diff
@@ -39,6 +39,16 @@ module.exports = {
       }
     ],
     "no-plusplus": "off",
-  }
+  },
+  overrides: [
+    {
+      "files": ["packages/theme/tests/e2e/**/*"],
+      "rules": {
+        "jest/expect-expect": "off",
```
sethidden, line 47: cypress uses similar test case syntax to jest so eslint gets confused and treats cypress like jest. this solution is not optimal but will do for now

packages/theme/package.json
```diff
@@ -41,8 +41,10 @@
     "@vue-storefront/nuxt": "~2.5.6",
     "@vue-storefront/redis-cache": "^1.0.1",
     "axios": "^0.26.1",
+    "body-parser": "1.19.2",
```
sethidden, line 44: These were already used by body-parser.js - I'm just adding them explicitly, rather than using the transient versions provided by unrelated packages

packages/theme/tests/e2e/pages/checkout/Billing.ts
```diff
@@ -0,0 +1,15 @@
+import { el } from '../utils/element';
```
sethidden, line 1: Split one file with 4 classes into 4 files with 1 class each (eslint rule complaining)

bartoszherba: Good :) This is how it should be done, one file - one class.

bartoszherba (changes requested):
@sethidden Please, fix build issues.

bartoszherba (approved):

PR #1009, opened 2022-05-17
fix: coupon code invalid error message

bartoszherba:
## Description
- rework coupon code component to handle GraphQL errors

## Related Issue
#957 

## How Has This Been Tested?

- add the product to the cart
- open the cart
- try to apply invalid coupon -> observe there are no errors in the console and the error message is verbose
- try to apply valid coupon -> observe that coupon was applied
- try to remove coupon -> observe the coupon was applied

Screenshot 2022-05-17 at 11 31 45

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/packages/theme/modules/checkout/composables/useCart/index.ts
+++ b/packages/theme/modules/checkout/composables/useCart/index.ts
@@ -197,21 +197,35 @@ PRODUCT
     }
   };
 
+  const handleCoupon = async (couponCode = null, customQuery = null): Promise<void> => {
+    const variables = {
+      currentCart: cart.value,
+      customQuery,
+      couponCode,
+    };
+
+    const { updatedCart, errors } = couponCode
+      ? await applyCouponCommand.execute(context, variables)
+      : await removeCouponCommand.execute(context, variables);
+
+    if (errors) {
+      throw errors[0];
+    }
+
+    if (updatedCart) {
+      customerStore.$patch((state) => {
+        state.cart = updatedCart;
+      });
+    }
+  };
+
   const applyCoupon = async ({ couponCode, customQuery }): Promise<void> => {
     Logger.debug('useCart.applyCoupon');
 
     try {
       loading.value = true;
-      const { updatedCart } = await applyCouponCommand.execute(context, {
-        currentCart: cart.value,
-        couponCode,
-        customQuery,
-      });
-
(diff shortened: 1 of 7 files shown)


packages/theme/modules/checkout/composables/useCart/index.ts
```diff
@@ -197,21 +197,40 @@ PRODUCT
     }
   };
 
-  const applyCoupon = async ({ couponCode, customQuery }): Promise<void> => {
-    Logger.debug('useCart.applyCoupon');
+  const handleCoupon = async (couponCode = null, customQuery = null): Promise<boolean> => {
```
sethidden, line 200:
```suggestion
  const handleCoupon = async (couponCode = null, customQuery = null): Promise<void> => {
```
What do you think about throwing an exception from handleCoupon if there are errors instead of returning false? You could then handle the exception with try/catch in applyCoupon

To me, a function succeeded if executed fully without throwing an excception, not by returning `true`

packages/theme/modules/checkout/composables/useCart/index.ts
```diff
@@ -197,21 +197,40 @@ PRODUCT
...
-        couponCode,
-        customQuery,
-      });
+    const { updatedCart, errors } = couponCode
+      ? await applyCouponCommand.execute(context, variables)
+      : await removeCouponCommand.execute(context, variables);
 
-      error.value.applyCoupon = null;
+    if (errors) {
+      [error.value.applyCoupon] = errors;
```
sethidden, line 212:
```suggestion
      error.value.applyCoupon = errors;
```
Not sure if my suggestion is correct. Can't wrap my head around whether this is detructuring?

sethidden (changes requested):
Left some comments

packages/theme/modules/checkout/composables/useCart/index.ts
bartoszherba: Because I want to get the first element. Apply coupon is not an array and I pretty much doubt that there will be more than one error in the same time..

packages/theme/modules/checkout/composables/useCart/index.ts
bartoszherba: ok

sethidden (approved):

PR #1014, opened 2022-05-18
refactor: create renderers for each product type

bartoszherba:
## Description
Currently, we are handling logic for any type of product directly inside the page component which is messy, hard to maintain, modify and understand. Because each product in Magento has its own features, we should add dedicated renderers with dedicated logic

refactor: create renderers for each product type
    - split root product page into a separate components
    - move styles to shared file
    - reduce the amount of unused code in different component types
    - add useGallery composable
    - add useProductTabs composable
    - add canAddToCart method on useCart composable

## Related Issue
#804 

## Motivation and Context
Improvement

Screenshot 2022-05-20 at 11 38 40

## Types of changes

## Checklist:


packages/theme/modules/catalog/product/components/tabs/ProductTabs.vue
```diff
@@ -0,0 +1,183 @@
...
+  components: {
+    ProductAddReviewForm,
+    HTMLContent,
+    SfReview,
+    SfTabs,
+    SfLoader,
+  },
+  props: {
+    product: {
+      type: [Object, null] as PropType<Product>,
```
sethidden, line 94:
```suggestion
      type: [Object, null] as PropType<Product | null>,
```

packages/theme/modules/catalog/product/composables/useGallery.ts
```diff
@@ -0,0 +1,21 @@
+import { computed } from '@nuxtjs/composition-api';
+import { getGallery as getProductGallery } from '~/modules/catalog/product/getters/productGetters';
+import { useImage } from '~/composables';
+import { Product } from '~/modules/catalog/product/types';
+
+export function useGallery(product: Product) {
```
sethidden, line 6:
```suggestion
export function useGallery(product: Ref<Product>) {
```

Is this reactive? I remember when I was creating useProductsWithCommonCardProps.ts I needed to pass a ref through the argument, not a regular value

packages/theme/modules/catalog/product/components/types/bundle/BundleProduct.vue
```diff
@@ -0,0 +1,221 @@
+<template>
```
sethidden, line 1: components/types could be confusing with the `types/` folders that we use for TypeScript types.

Maybe components/product-variants or just variants, idk

packages/theme/modules/catalog/product/components/tabs/ProductTabs.vue
```diff
@@ -0,0 +1,183 @@
...
+      return tabNumber;
+    };
+
+    const activeTab = computed(() => changeTab(props.openTab));
+
+    const productDescription = computed(
+      () => props.product?.description?.html || '',
+    );
+
+    const successAddReview = async (reviewData) => {
```
sethidden, line 159: Missing type - ProductReview may fit

packages/theme/modules/catalog/product/components/tabs/ProductTabs.vue
```diff
@@ -0,0 +1,183 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      await addReview(reviewData);
+      document.querySelector('#tabs').scrollIntoView({
```
sethidden, line 162: Why not use a ref rather than making the browser look for it in the DOM from scratch?

packages/theme/modules/catalog/product/components/tabs/ProductTabs.vue
```diff
@@ -0,0 +1,183 @@
...
+    <SfTab title="Description">
+      <HTMLContent
+        v-if="productDescription"
+        :content="productDescription"
+        tag="div"
+        class="product__description"
+      />
+    </SfTab>
+    <SfTab title="Read reviews">
+      <div v-show="isReviewsLoading">
```
sethidden, line 17: Why the v-show? Won't v-if with v-else be more readable? The review and loader block aren't that hughe to use display: none instead of re-rendering with v-if

packages/theme/modules/catalog/product/components/types/styles.scss
```diff
@@ -0,0 +1,189 @@
+.product {
```
sethidden, line 1: I assume this is going to be broken down later but it's pretty scary that all the product variant components use the same styles. Like, if I change one thing I'd be scared of the collateral damage leaking into other components.

But as I said, I assume you're going to be extracting certain shared things to separate components

sethidden (commented):
Given you've mentioned you still want to work on this in future iterations, don't you want to make a separate feature branch? I'm reluctant to merge this to develop directly because of all the duplication in template duplication between product variants

(not requesting changes or approving for now)

packages/theme/modules/catalog/product/components/tabs/ProductTabs.vue
bartoszherba: Fixed.

packages/theme/modules/catalog/product/composables/useGallery.ts
bartoszherba: I don't understand ;) this is MY NEW composable ;) Param is not a ref but the output is a computed value (at least partially).

packages/theme/modules/catalog/product/components/types/bundle/BundleProduct.vue
bartoszherba: -> product-types (type is a magento naming convention)

packages/theme/modules/catalog/product/components/tabs/ProductTabs.vue
bartoszherba: No, but I will add proper type ;)

packages/theme/modules/catalog/product/components/tabs/ProductTabs.vue
bartoszherba: Legacy - I created a task to fix that.

packages/theme/modules/catalog/product/components/tabs/ProductTabs.vue
bartoszherba: Fixed.

packages/theme/modules/catalog/product/components/types/styles.scss
bartoszherba: Do you suggest that I should move it back to components? Or maybe add style file for each group?

sethidden: > Do you suggest that I should move it back to components?

I mean move it to even more deeply nested components, so the component tree would look like

product.vue > BundledProduct.vue > ProductReviews.vue
product.vue > ConfigurableProduct.vue > ProductReviews.vue

etc.

for more strict encapsulation in a component instead of sharing styles

bartoszherba: Ok, I will create a task for that too.

sethidden (dismissed):

sethidden (approved):

PR #1016, opened 2022-05-18
refactor: remove useless order getters

ghost:
Removed order getters that don't really do anything.

The ones I left either have:
* some nontrivial calculation
* some value in the form of pointing to which property you should use. like, orders have many types of prices, so it's good to have a single source of truth that points to that "one true" property to use

### DIFF
--- a/packages/theme/modules/checkout/getters/orderGetters.ts
+++ b/packages/theme/modules/checkout/getters/orderGetters.ts
@@ -1,13 +1,11 @@
-import { AgnosticPagination } from '~/composables/types';
-import { CustomerOrders, CustomerOrder, OrderItemInterface } from '~/modules/GraphQL/types';
+import type { AgnosticPagination } from '~/composables/types';
+import type { CustomerOrders, CustomerOrder, OrderItemInterface } from '~/modules/GraphQL/types';
 
-export const getDate = (order: CustomerOrder): string => new Date(order?.order_date?.replace(/ /g, 'T')).toLocaleDateString() || '';
+export const getDate = (order: CustomerOrder): string => new Date(order?.order_date?.replace(/ /g, 'T')).toLocaleDateString();
 
-export const getStatus = (order: CustomerOrder): string => order?.status || 'Failed';
+export const getPrice = (order: CustomerOrder): number => order?.total?.base_grand_total?.value ?? 0;
 
-export const getPrice = (order: CustomerOrder): number | null => order?.total?.base_grand_total?.value || 0;
-
-export const getItemPrice = (item: OrderItemInterface): number => item?.product_sale_price?.value || 0;
+export const getItemPrice = (item: OrderItemInterface): number => item?.product_sale_price?.value ?? 0;
 
 const getPagination = (orders: CustomerOrders): AgnosticPagination => ({
   currentPage: orders?.page_info?.current_page || 1,
@@ -19,7 +17,6 @@ const getPagination = (orders: CustomerOrders): AgnosticPagination => ({
 
 const orderGetters = {
   getDate,
-  getStatus,
   getPrice,
   getItemPrice,
   getPagination,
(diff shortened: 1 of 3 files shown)


packages/theme/modules/checkout/getters/orderGetters.ts
```diff
@@ -1,27 +1,17 @@
...
-export const getItemName = (item: OrderItemInterface): string => item?.product_name || '';
-
-export const getItemQty = (item: OrderItemInterface): number => item?.quantity_ordered || 0;
-
 export const getItemPrice = (item: OrderItemInterface): number => item?.product_sale_price?.value || 0;
 
-export const getFormattedPrice = (price: OrderItemInterface) => String(price);
-
-const getPagination = (orders: any): AgnosticPagination => ({
+const getPagination = (orders: CustomerOrders): AgnosticPagination => ({
```
sethidden, line 14: removed any

packages/theme/modules/checkout/getters/orderGetters.ts
```diff
@@ -1,27 +1,17 @@
 import { AgnosticPagination } from '~/composables/types';
-import { CustomerOrder, OrderItemInterface } from '~/modules/GraphQL/types';
+import { CustomerOrders, CustomerOrder, OrderItemInterface } from '~/modules/GraphQL/types';
 
 export const getDate = (order: CustomerOrder): string => new Date(order?.order_date?.replace(/ /g, 'T')).toLocaleDateString() || '';
 
-export const getId = (order: CustomerOrder): string => order?.number || '';
-
 export const getStatus = (order: CustomerOrder): string => order?.status || 'Failed';
```
bartoszherba, line 6: Please, "failed" is not a good status, change it maybe to `unknown` or something that will mark it in a similar way.

packages/theme/modules/checkout/getters/orderGetters.ts
```diff
@@ -1,27 +1,17 @@
...
 
 export const getPrice = (order: CustomerOrder): number | null => order?.total?.base_grand_total?.value || 0;
```
bartoszherba, line 8:
```suggestion
export const getPrice = (order: CustomerOrder): number => order?.total?.base_grand_total?.value ?? 0;
```

packages/theme/modules/checkout/getters/orderGetters.ts
```diff
@@ -1,27 +1,17 @@
 import { AgnosticPagination } from '~/composables/types';
-import { CustomerOrder, OrderItemInterface } from '~/modules/GraphQL/types';
+import { CustomerOrders, CustomerOrder, OrderItemInterface } from '~/modules/GraphQL/types';
```
bartoszherba, line 2:
```suggestion
import type { CustomerOrders, CustomerOrder, OrderItemInterface } from '~/modules/GraphQL/types';
```

packages/theme/modules/checkout/getters/orderGetters.ts
```diff
@@ -1,27 +1,17 @@
...
 
 export const getDate = (order: CustomerOrder): string => new Date(order?.order_date?.replace(/ /g, 'T')).toLocaleDateString() || '';
```
bartoszherba, line 4:
```suggestion
export const getDate = (order: CustomerOrder): string => new Date(order?.order_date?.replace(/ /g, 'T')).toLocaleDateString();
```

bartoszherba (changes requested):

packages/theme/modules/checkout/getters/orderGetters.ts
sethidden: Removed geStatus entirely 8bb720ab4fd6afea392cd0c02b37c8a02a1762c8

packages/theme/modules/checkout/getters/orderGetters.ts
sethidden: 8bb720ab4fd6afea392cd0c02b37c8a02a1762c8

packages/theme/modules/checkout/getters/orderGetters.ts
sethidden: 8bb720ab4fd6afea392cd0c02b37c8a02a1762c8

packages/theme/modules/checkout/getters/orderGetters.ts
sethidden: 8bb720ab4fd6afea392cd0c02b37c8a02a1762c8

bartoszherba (approved):

PR #1023, opened 2022-05-19
chore: implement overlooked additional tasks

ghost:
In M2-549 and M2-547 there were some additional tasks given - removal of "Returns" tab and removal of list view from my wishlist - that I overlooked.


sethidden:
@bartoszherba 
Please note I removed the whole top header aswell. Without the "list/grid" view buttons the header with the product count just looked weird. I can revert it easily since it's in a separate commit, so let me know what do you think

image
image

bartoszherba (approved):

PR #1026, opened 2022-05-19
feat: add product to cart from wishlist

AlexanderDevitsky:
## Related Issue

https://vsf.atlassian.net/browse/M2-502

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/modules/catalog/pages/product.vue
+++ b/packages/theme/modules/catalog/pages/product.vue
@@ -83,9 +83,11 @@ export default defineComponent({
           eq: id,
         },
       },
-      configurations: Object.entries(route.value.query).map(
-        (config) => config[1],
-      ) as string[],
+      configurations: Object.entries(route.value.query)
+        .filter((config) => config[0] !== 'wishlist')
+        .map(
+          (config) => config[1],
+        ) as string[],
     });
 
     // eslint-disable-next-line no-underscore-dangle
@@ -107,7 +109,7 @@ export default defineComponent({
       const tags = [
         {
           prefix: CacheTagPrefix.View,
-          value: `product-${route.value.params.id}`,
+          value: `product-${id}`,
         },
       ];
 
--- a/packages/theme/modules/customer/pages/MyAccount/MyWishlist.vue
+++ b/packages/theme/modules/customer/pages/MyAccount/MyWishlist.vue
@@ -143,7 +143,7 @@ import wishlistGetters from '~/modules/wishlist/getters/wishlistGetters';
 import { useCart } from '~/modules/checkout/composables/useCart';
 import { useWishlistStore } from '~/modules/wishlist/store/wishlistStore';
 import EmptyWishlist from '~/modules/wishlist/components/EmptyWishlist.vue';
-
+import { ProductTypeEnum } from '~/modules/catalog/product/enums/ProductTypeEnum';
 import { useUiHelpers, useImage } from '~/composables';
 
 export default defineComponent({
@@ -187,7 +187,7 @@ export default defineComponent({
       const productType = product.__typename;
 
       switch (productType) {
-        case 'SimpleProduct':
+        case ProductTypeEnum.SIMPLE_PRODUCT:
           await addItemToCartBase({
             product,
             quantity,
@@ -197,8 +197,9 @@ export default defineComponent({
             cartError: cartError.value.addItem,
           });
           break;
-        case 'BundleProduct':
-        case 'ConfigurableProduct':
+        case ProductTypeEnum.CONFIGURABLE_PRODUCT:
+        case ProductTypeEnum.BUNDLE_PRODUCT:
+        case ProductTypeEnum.GROUPED_PRODUCT:
           const path = `/p/${productGetters.getProductSku(
             product,
           )}${productGetters.getSlug(product, product.categories[0])}`;
--- a/packages/theme/modules/customer/pages/MyAccount/MyWishlist.vue
+++ b/packages/theme/modules/customer/pages/MyAccount/MyWishlist.vue
@@ -68,6 +68,7 @@
                   wishlist-icon
                   is-in-wishlist
                   show-add-to-cart-button
+                  :add-to-cart-disabled="isCartLoading"
                   @click:wishlist="removeItemFromWishlist(product.product)"
                   @click:add-to-cart="
                     addItemToCart({ product: product.product, quantity: 1 })
@@ -158,16 +159,24 @@ export default defineComponent({
   },
   setup() {
     const {
-      load, loading, removeItem,
+      load,
+      loading,
+      removeItem,
+      afterAddingWishlistItemToCart,
     } = useWishlist();
     const route = useRoute();
-    const { app } = useContext();
+    const { localeRoute } = useContext();
     const {
       query: { page, itemsPerPage },
     } = route.value;
     const router = useRouter();
     const th = useUiHelpers();
-    const { addItem: addItemToCartBase, isInCart } = useCart();
+    const {
+      addItem: addItemToCartBase,
+      error: cartError,
+      isInCart,
+      loading: isCartLoading,
+    } = useCart();
     const wishlistStore = useWishlistStore();
 
     const products = computed(() => wishlistGetters.getProducts(wishlistStore.wishlist));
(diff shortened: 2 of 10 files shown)


AlexanderDevitsky:
@bartoszherba should I create unit-tests in the scope of this task?

bartoszherba:
@AlexanderDevitsky I must put this on hold for a while, if you look at this draft https://github.com/vuestorefront/magento2/pull/1014 you will realize why. There will be multiple problems like we no longer can have business logic on the product page. I also do not like the fact that we base on the locale path as the solution is not generic, eg. if someone will add a wishlist somewhere else, or create a different path for a subaccount, or [name it] it will not work. I need to think of how it can be solved. For now we must hold this PR ;(

packages/theme/modules/catalog/pages/product.vue
```diff
@@ -560,9 +572,48 @@ export default defineComponent({
...
+        && isProductInWishlist.value
+        && !cartError.value?.addItem
+      ) {
+        await removeItemFromWishlist({ product: product.value });
+
+        sendNotification({
+          id: Symbol('product_added_to_cart_from_wishlist_product_page'),
+          message: i18n.t(`You added ${getProductName(
+            product.value,
+          )} to your shopping cart.`) as string,
```
sethidden, line 590: Please pass product name as parameter to the label. We can't possibly make thousands of translation labels for each product name

https://kazupon.github.io/vue-i18n/guide/formatting.html#named-formatting

```suggestion
          message: i18n.t(`You added productname to your shopping cart.`, { productname: getProductName(product.value)} ) as string,
```

This won't render the product name after this change obviously, so you'll need to add that to the translations in en.js, de.js like `'You added productname to your shopping cart': 'You added {productname} to your shopping cart'`.

Don't forget to do that in MyWishlist aswell.

@bartoszherba is it ok to add new translation labels like that?

bartoszherba: No, the product's translation comes from Magento. We mustn't translate it on the front. Good catch, thanks.

packages/theme/modules/catalog/pages/product.vue
```diff
@@ -560,9 +572,48 @@ export default defineComponent({
...
+          type: 'danger',
+          icon: 'cross',
+          persist: false,
+          title: 'Wishlist error',
+        });
+      }
+    };
+
+    onMounted(() => {
+      if (from.value?.path === localePath('/customer/my-wishlist')) {
```
sethidden, line 609: 1. Wondering if it wouldn't be easier to just redirect to the product page with some query parameter `/p/product?from-wishlist` - that way if someone refreshes the page the "fromWishlist" information won't be lost

2. Shouldn't this be a computed?

3. If the query paramater doesn't work, it'll be easier to maybe use the route *name*, not the route path

```suggestion
      if (from.value?.name === localeRoute({ name: 'customer-my-wishlist' })) {
```

(see modules/customer/index.ts for route name)

packages/theme/modules/customer/pages/MyAccount/MyWishlist.vue
```diff
@@ -316,6 +320,29 @@ export default defineComponent({
...
+              id: Symbol('product_added_to_cart_from_wishlist_wishlist_page'),
+              message: app.i18n.t(`You added ${productGetters.getName(
+                product,
+              )} to your shopping cart.`) as string,
+              type: 'success',
+              icon: 'check',
+              persist: false,
+              title: 'Wishlist',
+            });
+          } else if (cartError.value.addItem.message) {
```
sethidden, line 336:
```suggestion
          } else {
```

would that work? I don't think errors can have missing messages

sethidden (changes requested):

AlexanderDevitsky:
@bartoszherba In this case, we need to extend Magento GraphQL schema by adding a 'remove_from_wishlist' field. Then implement the logic of removing a product from the wishlist on the Magento side. As I understand it, for the default theme, we cannot use additional implementations for Magento, but must use only the default implementation. Therefore, this option is not possible.
Currently, the referrer url is only used to remove configurable product when navigating to it from the wishlist.
If a product needs to be removed from the wishlist anyway, not just when added from the wishlist, then we just need to check if it's on the wishlist and using referrer url is redundant.
When https://github.com/vuestorefront/magento2/pull/1014 is finished, let me know and we will find a better solution for this.

packages/theme/modules/customer/pages/MyAccount/MyWishlist.vue
AlexanderDevitsky: @sethidden I think the message should be there anyway, so this can be removed

bartoszherba:
@AlexanderDevitsky 
PR is merged, I am sorry for that inconvenience.

With regards to the solution:

I think that we can make it work as @sethidden suggested. I did a small POC and reworked the query build in `product.vue` to something like, for a query parameter `wishlist=true` => example `http://localhost:3000/default/p/WJ01/stellar-solar-jacket.html?wishlist=true&OTM%3D=Y29uZmlndXJhYmxlLzkzLzYw&MTQ0=Y29uZmlndXJhYmxlLzE0NC8xNjk%3D`
```
const getBaseSearchQuery = () => ({
      filter: {
        sku: {
          eq: id,
        },
      },
      configurations: Object.entries(route.value.query)
        .filter((config) => config[0] !== 'wishlist')
        .map(
          (config) => config[1],
        ) as string[],
    });
```

Now, we can trigger wishlist mode on any PDP page with this flag from any place simply by passing the query param and we are not tight to the account page. If you look on the Magento, there is something very similar, as the add-to-wishlist has its own controller.

That would be the first step.

The second step is to have generic and reusable addToWishlist logic, separated from PDP (unlike you did in your PR which is justified by how it looked at that time).
Important thing is to keep business logic out of components, especially out of the page. This is not completely done at this moment because I did not want to keep such a huge change any longer, but this is the goal, so by introducing a new feature we can do it right from the very beginning.

@sethidden @AlexanderDevitsky What do you think?

AlexanderDevitsky:
@bartoszherba Good solution

bartoszherba (changes requested):
Please, review all comments.

AlexanderDevitsky:
@bartoszherba @sethidden I have added new changes. Please review if this is the right solution for us

packages/theme/modules/customer/pages/MyAccount/MyWishlist.vue
```diff
@@ -183,13 +192,22 @@ export default defineComponent({
             product,
             quantity,
           });
+          afterAddingWishlistItemToCart({
+            product,
+            cartError: cartError.value.addItem,
+          });
           break;
         case 'BundleProduct':
```
bartoszherba, line 200: could you please add GroupedProduct? I know that it is not your change but I just noticed that. Also, you can replace it with `packages/theme/modules/catalog/product/enums/ProductTypeEnum.ts`

packages/theme/modules/wishlist/composables/useWishlist/index.ts
```diff
@@ -10,7 +10,10 @@ import type {
   UseWishlistIsInWishlistParams,
   UseWishlistLoadParams,
   UseWishlistRemoveItemParams,
+  UseWishlistAfterAddingWishlistItemToCartParams,
 } from '~/modules/wishlist/composables/useWishlist/useWishlist';
+import { useUiNotification } from '~/composables';
+import productGetters from '~/modules/catalog/product/getters/productGetters';
```
bartoszherba, line 16: Could you please remove that and simply `product.name` in the code? We will remove most of the getters and we avoid using getters in new code unless it is necessary, eg. when getter execute some additional logic.

bartoszherba (changes requested):
@AlexanderDevitsky 
Thank you for changes, just last 2 small comments and we are done here ;)

packages/theme/modules/customer/pages/MyAccount/MyWishlist.vue
AlexanderDevitsky: done

packages/theme/modules/wishlist/composables/useWishlist/index.ts
AlexanderDevitsky: done

bartoszherba (approved):

bartoszherba (approved):

Frodigo (approved):

sethidden (approved):

PR #1027, opened 2022-05-19
fix: prevent SfSidebar disableBodyScroll triggering on desktop

ghost:
As part of https://github.com/vuestorefront/storefront-ui/pull/619
body-scroll-lock was added to SfSidebar.
If you look closely at
https://github.com/vuestorefront/storefront-ui/blob/d5b646a6344e0ab4f5dad214458662af7c8bbd57/packages/vue/src/components/organisms/SfSidebar/SfSidebar.vue#L146
 
you'll see that disableBodyScroll is called in a watcher on the "visible" prop.
hint: disableBodyScroll() just sets "overflow: hidden" on the HTML
document body, which makes the body unscrollable.
In this case that I'm fixing, it goes like:

    The value passed to CategoryFilters's isVisible comes from
    useUiState's isFilterSidebarOpen. This value is false by default.

    Clicking the "clear" or "apply" button in the filters emits the
    "close" event from CategoryFilters - if you look at category.vue that
    just calls "toggleFilterSidebar". This will set isVisible to TRUE

    Changing isVisible to true triggers SfSidebar's watcher, even though the SfSidebar
    it is not currently visible (it should be visible only on mobile)

    This in turn runs disableBodyLock() and prevents scrolling

### DIFF
--- a/packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
+++ b/packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
@@ -63,8 +63,10 @@
         </SfButton>
       </div>
     </div>
+    <!-- Do not use :visible="isVisible" - it causes M2-571 -->
     <SfSidebar
-      :visible="isVisible"
+      v-if="isVisible"
+      visible
       class="sidebar-filters smartphone-only"
       title="Filters"
       @close="$emit('close')"


bartoszherba:
@sethidden As long as it works.... :)

bartoszherba (approved):

packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
```diff
@@ -63,8 +63,10 @@
         </SfButton>
       </div>
     </div>
+    <!-- Do not use :visible="isVisible" - it causes M2-571 -->
```
bartoszherba, line 66: Could you please create an issue for a SFUI and link it here as a TODO + link to the issue?

sethidden: (after speaking on slack - this probably isn't a case sfui is willing to support)

PR #1034, opened 2022-05-20
perf: remove TopBar layout shift

ghost:
See https://vsf.atlassian.net/browse/M2-611

### DIFF
--- a/packages/theme/components/TopBar/TopBar.vue
+++ b/packages/theme/components/TopBar/TopBar.vue
@@ -18,26 +18,27 @@
   </SfTopBar>
 </template>
 
-<script>
+<script lang="ts">
+import { defineComponent } from '@nuxtjs/composition-api';
 import { SfButton, SfTopBar } from '@storefront-ui/vue';
-import useTopBar from './useTopBar';
+import { useTopBar } from './useTopBar';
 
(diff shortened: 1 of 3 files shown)


packages/theme/components/TopBar/TopBar.vue
```diff
@@ -18,26 +18,27 @@
...
-    CurrencySelector: () => import('../CurrencySelector'),
+    CurrencySelector: () => import('../CurrencySelector.vue'),
+    StoreSwitcher: () => import('../StoreSwitcher.vue'),
     SfTopBar,
     SfButton,
-    StoreSwitcher: () => import('../StoreSwitcher'),
   },
   setup() {
-    const { hasStoresToSelect, hasCurrencyToSelect } = useTopBar();
+    const { hasCurrencyToSelect, hasStoresToSelect } = useTopBar();
```
bartoszherba, line 34: Thanks for that change, it was important ;)

bartoszherba (approved):

packages/theme/components/TopBar/TopBar.vue
sethidden: It's in the order it appears in the template :P

PR #1036, opened 2022-05-20
refactor: fix eslint warnings in .vue files

sethidden:
Continued fixes of eslint, but this time in .vue files.

Turned out that a huge amount of .vue errors were in a file that contained some SVG tags.

AFter this PR, there are 30 errors left - most are about missing lang="ts"


sethidden:
Added lang="ts" to some more components. There are 17 errors left. Most are lang="ts", but adding lang="ts" obviously makes that one error turn into seven.

packages/theme/pages/Page.vue
```diff
@@ -41,22 +38,19 @@ export default defineComponent({
...
-    } = useContent('cmsPage');
+    const { loadPage, error, loading } = useContent();
 
     const route = useRoute();
     const { error: nuxtError, app } = useContext();
     const { params } = route.value;
-    const page = ref({});
+    const page = ref<CmsPage | null>(null);
 
+    // eslint-disable-next-line consistent-return
```
bartoszherba, line 48: Is this still required? maybe add return value instead of suppressing linter.

packages/theme/pages/Page.vue
```diff
@@ -71,13 +65,16 @@ export default defineComponent({
       return null;
     }
 
-    const title = this.page?.meta_title ? this.page.meta_title : this.page.title;
+    // eslint-disable-next-line prefer-destructuring
+    const page : CmsPage = this.page;
```
bartoszherba, line 69:
```suggestion
    const { page }: CmsPage = this;
```

packages/theme/pages/Page.vue
```diff
@@ -14,17 +14,14 @@
...
-} from '@storefront-ui/vue';
+<script lang="ts">
+import { SfLoader, SfHeading } from '@storefront-ui/vue';
 import {
   defineComponent, ref, useContext, useFetch, useRoute,
 } from '@nuxtjs/composition-api';
 import { useCache, CacheTagPrefix } from '@vue-storefront/cache';
-import { useContent } from '~/composables';
-import HTMLContent from '~/components/HTMLContent';
+import { CmsPage, useContent } from '~/composables';
```
bartoszherba, line 23: Please, import CmsPage from GraphQL module.

packages/theme/pages/Page.vue
bartoszherba: Anyway, there is something wrong with this type if you destruct the value, is the page really CmsPage?

bartoszherba (changes requested):

packages/theme/pages/Page.vue
sethidden: https://nuxtjs.org/docs/directory-structure/middleware/#named-middleware

Given the example above I don't think there's anything to return. If I add just `return` the linter will complain about "useless return"

packages/theme/pages/Page.vue
sethidden: Avoided destructuring with type assertion by using a solution from nuxt/typescript (link in commit message)

Frodigo (approved):

bartoszherba (approved):

PR #1059, opened 2022-05-25
fix: disable body scroll lock

ghost:
Previous PR got back from QA as not fixing the issue. Have to use the uglier workaround - SfSidebar doesn't block scrolling anyway (double scrollbars on stage in filter sidebar), so I'm not worried about disabling the body scroll lock globally.

### DIFF
--- a/packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
+++ b/packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
@@ -63,10 +63,8 @@
         </SfButton>
       </div>
     </div>
-    <!-- Do not use :visible="isVisible" - it causes M2-571 -->
     <SfSidebar
-      v-if="isVisible"
-      visible
+      :visible="isVisible"
       class="sidebar-filters smartphone-only"
       title="Filters"
       @close="$emit('close')"
@@ -114,7 +112,7 @@
 </template>
 <script lang="ts">
 import {
-  defineComponent, onMounted, provide, Ref, ref,
+  defineComponent, onMounted, provide, Ref, ref, nextTick, watch,
 } from '@nuxtjs/composition-api';
 import {
   SfAccordion,
@@ -125,6 +123,7 @@ import {
   SfSidebar,
 } from '@storefront-ui/vue';
 
+import { clearAllBodyScrollLocks } from 'body-scroll-lock';
 import SkeletonLoader from '~/components/SkeletonLoader/index.vue';
 import { useUiHelpers } from '~/composables';
 import { getFilterConfig, getDisabledFilters } from '~/modules/catalog/category/config/FiltersConfig';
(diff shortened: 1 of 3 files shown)


packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
```diff
@@ -188,6 +189,7 @@ export default defineComponent({
       }
       emit('reloadProducts');
       emit('close');
+      document.querySelector(categoryNavbarIdSelector).scrollIntoView();
```
bartoszherba, line 192: Is this necessary? Can't we use refs?

packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
```diff
@@ -204,8 +207,17 @@ export default defineComponent({
       updateRemovableFilters();
       emit('reloadProducts');
       emit('close');
+      document.querySelector(categoryNavbarIdSelector).scrollIntoView();
     };
 
+    watch(() => props.isVisible, (newValue) => {
+      // disable Storefrontt UI's body scroll lock which is launched when :visible prop on SfSidebar changes
```
bartoszherba, line 214: Can we create a task to get rid of the SfSidebar?

bartoszherba (dismissed):

packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
Frodigo: We cannot manipulate other components in this way. One component cannot secretly change the other.

packages/theme/package.json
```diff
@@ -42,6 +42,7 @@
     "@vue-storefront/redis-cache": "^1.0.1",
     "axios": "^0.26.1",
     "body-parser": "1.19.2",
+    "body-scroll-lock": "^3.0.1",
```
Frodigo, line 45: Do we really need a new dependency? This one was lastly published a year ago. It doesn't look like something actively maintained.

RetroModernDev (changes requested):

packages/theme/package.json
sethidden: This dependency is part of storefront ui https://github.com/vuestorefront/storefront-ui/blob/d5b646a6344e0ab4f5dad214458662af7c8bbd57/packages/vue/src/components/organisms/SfSidebar/SfSidebar.vue#L65

We already use it vicariously through sfui, I'm adding it with the same version as in sfui here because I can't import something that's not in package.json (eslint rule). The bundle size should stay the same.

packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
sethidden: [`5a49847` (#1059)](https://github.com/vuestorefront/magento2/pull/1059/commits/5a49847fd8d30f1b7642105e9d84bc2b1ce0ecee)

packages/theme/modules/catalog/category/components/filters/CategoryFilters.vue
sethidden: https://vsf.atlassian.net/browse/M2-688

bartoszherba (approved):

RetroModernDev (approved):

PR #1066, opened 2022-05-26
refactor: remove some instances of implicit any

sethidden:
Note this PR is meant to remove only **implicit** any, such as writing the below in a .ts file

```ts
const func = (a) => // a is any, even though I didn't write a: any
```

I tried setting noImplicitAny in tsconfig so the cases I'm removing here wouldn't even build but there's too many cases to fix. Even after this PR, there are ~140 implicit any related errors left.

I only fixed the easy ones here. The ones that are left are more difficult (require changes in implementation, removal of certain files etc.)


packages/theme/helpers/integrationPlugin/context.ts
```diff
@@ -1,29 +1,35 @@
...
+
 /**
  * It extends given integartion, defined by `tag` in the context.
  */
-export const createExtendIntegrationInCtx = ({ tag, nuxtCtx, inject }) => (integrationProperties: Record<string, unknown>) => {
+export const createExtendIntegrationInCtx = ({ tag, nuxtCtx, inject } : Argument) => (integrationProperties: Record<string, unknown>) => {
   const integrationKey = `$${tag}`;
 
-  if (!nuxtCtx.$vsf || !nuxtCtx.$vsf[integrationKey]) {
+  if (!nuxtCtx.$vsf || !(nuxtCtx.$vsf as Record<string, any>)[integrationKey]) {
```
sethidden, line 14: Intentional cast with any since it's expected that you can assign anything

packages/theme/modules/magento/defaultConfig.ts
```diff
@@ -8,7 +8,7 @@ export const defaultConfig = {
     storeCookieName: 'vsf-store',
     messageCookieName: 'vsf-message',
   },
-  locale: undefined,
-  country: undefined,
-  currency: undefined,
+  locale: undefined as undefined,
```
sethidden, line 11: Can't be avoided

packages/theme/types/core.ts
```diff
@@ -1,902 +1,15 @@
-/* istanbul ignore file */
```
sethidden, line 1: I asked about this file on Slack. I don't really see it being used apart from the few interfaces I left at the bottom.

packages/theme/modules/customer/composables/useUserAddress/commands/createCustomerAddressCommand.ts
```diff
@@ -1,8 +1,11 @@
+import type { useContext } from '@nuxtjs/composition-api';
 import { CustomerAddressInput } from '~/modules/GraphQL/types';
 
+type Context = ReturnType<typeof useContext>;
```
sethidden, line 4: Should probably move this to a file rather than duplicating it 20 times across the repo...

sethidden: [`7af5dc0` (#1066)](https://github.com/vuestorefront/magento2/pull/1066/commits/7af5dc0907ced6120023ef6515d2ac25d55a9143)

packages/theme/modules/customer/helpers/generateUserData.ts
```diff
@@ -1,6 +1,6 @@
 import type { CustomerCreateInput } from '~/modules/GraphQL/types';
 
-export const generateUserData = (userData): CustomerCreateInput => {
+export const generateUserData = (userData: any): CustomerCreateInput => {
```
Frodigo, line 3: Have you tried with Customer or User instead of any here?

sethidden: Yes but they didn't fit. This is M2-664

Frodigo (approved):

PR #1068, opened 2022-05-27
refactor: use null instead of {}

sethidden:
In Checkout.vue and Billing.vue there was a case where `getSomeData() || {}` was used. This was probably done to avoid worrying if something is null or not and avoiding the "x is null, trying to access property abc" error.

The problem with that is that it also causes problems with typing, as you'd get errors like "X is not assignable to {}"


packages/theme/helpers/checkout/address.ts
```diff
@@ -0,0 +1,55 @@
+import type { CustomerAddress, CartAddressInterface } from '~/modules/GraphQL/types';
```
sethidden, line 1: (file was moved from address.js)

packages/theme/helpers/checkout/address.ts
```diff
@@ -0,0 +1,55 @@
...
+  street: string;
+  apartment: string;
+  city: string;
+  region: string;
+  country_code: string;
+  postcode: string;
+  telephone: string;
+}
+
+export const getInitialCheckoutAddressForm = () : CheckoutAddressForm => ({
```
sethidden, line 45: Thanks to this, the {} can be avoided

packages/theme/modules/checkout/getters/cartGetters.ts
```diff
@@ -140,9 +140,6 @@ export const getTotalItems = (cart: Cart): number => {
 
 export const getConfiguredVariant = (product: ConfigurableCartItem): ProductInterface | null => product?.configured_variant || null;
 
-// eslint-disable-next-line import/no-named-as-default-member
-export const getFormattedPrice = (price: number) => getFormattedPrice(price);
```
sethidden, line 144: removed weird self-referencing getter. it's not used anywhere

packages/theme/helpers/checkout/address.ts
Frodigo: Maybe the file should be located in modules/checkout/helpers?

Frodigo (approved):

Frodigo (approved):

PR #1072, opened 2022-05-27
docs: added documentation about global state management

Frodigo:
## Description
Documentation about global state management


docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
+# Global state management
+
+For global state management, we decided to use the [Pinia](https://pinia.vuejs.org/ssr/nuxt.html) library. We assume that only data needed in more than one place of the application should be kept globally. For example:
+
+- store config
+- categories that are used in navigation (different components for desktop and mobile)
```
sethidden, line 6:
```suggestion
- categories used in navigation (different components for desktop and mobile)
```
nitpick

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+- customer (user) data
+- cart/checkout data
+
+Data that is used in one place in the application should be stored locally in composables (as a local state)
```
sethidden, line 10:
```suggestion
Data that is used in one place in the application should be stored locally in composables
```
stored locally in composables as local state is as we say in Poland - buttery butter

```suggestion
Data that's not shared throughout the application should be stored locally in composables (as a local state)
```
"data that is used in one place in the application" - I think you meant data that's not shared?

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Difference between global and local state
+
+Sometimes you need to get some information in backend and display in on the screen. For example you get products from Magento and display it on category page. You need those products only in one place. So you need that data only locally in one place.
```
sethidden, line 14:
```suggestion
An example of local state is when you download product data from Magento and display it . You need those products only in that one view. So you should keep that data contained locally only to that view.
```

* "sometimes you need to get some"
* "category page" feels to me more like our internal lingo, unless this is  Magento thing
* rework "locally in one place"

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+
+On the other hand, some data should be accessible from more than one place in app. Examples:
```
sethidden, line 16:
```suggestion
On the other hand, some data should be accessible from more than one place in the app. Examples:
```

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+
+- cart - you need data about cart in minicart overlay, but also you need to access cart for example if you want to add product to the cart from category pager or product page.
```
sethidden, line 18:
```suggestion
- cart - you fetch cart data so you can show its contents in the the cart sidebar,  but also need cart data when  adding product to the cart from a product category page or a single product's page.
```

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+- configuration - various components can consume configuration so it’s also should be accessible globally
```
sethidden, line 19: This is too vague to me. A good example would be useUiState which we use to show globally available components (sidebars, etc. toasts?)

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+- navigation - menus are visible on desktop and mobile devices. They display for example top level categories, but mobile menu and desktop menu is totally different components. You can store information about categories globally and consume that information by any components
```
sethidden, line 20:
```suggestion
- navigation - because of screen size limitations, certain UI elements use the same data, but display it in a different way. For example, when browsing the product category tree in our integration, the mobile category menu is a full page overlay, while on the desktop it's embedded in the page's top header. The mobile and desktop components are far away from each other in the application's structure, but thanks to global state they can get data from the same source. 
```

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Pinia stores provided by Vue Storefront 2 for Magento integration
+
+Take a look at stores provided by the integration OOTB:
+
+- **Config store** - it provides information about available Magento stores, currency, and active store configuration
+- **Category store** - it provides information about categories (that is displayed in navigation)
```
sethidden, line 27:
```suggestion
- **Category store** - it provides information about product categories
```
specify which categories
I removed the (displayed in navigation) part since I thought it could be skipped

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+- **Customer store** - it provides information about customer, including information is customer logged-in or not
```
sethidden, line 28:
```suggestion
- **Customer store** - it provides information about the currently logged-in customer.  Initially, it contains a boolean flag that lets you know if there's a valid login session. If necessary, more detailed data (shipping, billing addresses etc.) are fetched.
```

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+- **Cart store** - it provides information about cart
```
sethidden, line 29: Could this be made more specific?

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+
+## How to create a Pinia store
+
+In following example, I define a new store named `useWishlistStore`
```
sethidden, line 33:
```suggestion
In the following example, I define a new store named `useWishlistStore`
```

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+
+```javascript
```
sethidden, line 35:
```suggestion
```typescript
```

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+});
+```
+
+The `useWishlistStore` has one state field: `wishlist` that is initially null.
+
+### Cosuming data from Pinia store
+
+Consuming data from the store is really straightforward. Basically you need to import it and use similarly as composable:
+
+```javascript
```
sethidden, line 56:
```suggestion
```typescript
```

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+import { useWishlistStore } from '~/stores/wishlistStore'
+
+export default {
```
sethidden, line 59: I assume you did this for simplicity but optionally yo could add defineComponent here

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+
+    return {
+      // you can return the whole store instance to use it in the template
+      wishlistStore
+    }
+  },
+}
+```
+
+In case if you want to return only **specific fields**, you need to wrap field with computed. If you don’t do that, reactivity will break.
```
sethidden, line 71:
```suggestion
In case you want to return only **specific fields**, you need to wrap the field with `computed()`. If you don’t do that, reactivity will break.
```

It's also possible to use storeToRefs to pluck properties

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+    const wishlist = computed(() => wishlistStore.wishlist);
+
+    return {
+      wishlist,
+    }
+  },
+}
+```
+
+### Modiyfying data in Pinia store
```
sethidden, line 88:
```suggestion
### Modifying data in Pinia store
```

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+
+To write data to the store we recommend using $patch like this:
+
+```javascript
+const { data } = await app.$vsf.$magento.api.wishlist();
+  wishlistStore.$patch((state) => {
+  state.wishlist = data.data?.customer?.wishlists ?? [];
```
sethidden, line 95: there's wrong indentation here

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+});
+```
+
+## How to use API in Pinia stores
+
+If you want to communicate with APi directly in store it’s possibly by using the $graphql nuxt plugin. To do so you need to define a new action in store, let’s define the `load` action:
```
sethidden, line 101:
```suggestion
If you want to communicate with an API directly in store it’s possibly by using the $graphql Nuxt plugin. To do so you need to define a new action in store, let’s define the `load` action:
```

sethidden, line 101: I'm not sure we want to recommend using $graphql - we have two API clients - graphql request and context.magento.api.$customQuery? Do we really want to commit to $graphql? I remember we wanted to remove it :P

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+  wishlist: Wishlist[] | null
+}
+
+export const useWishlistStore = defineStore('wishlist', {
+  state: (): WishlistStore => ({
+    wishlist: null,
+  }),
+  actions: {
+    async load() {
+      // communication with API will be here
```
sethidden, line 117: I think this code block isn't necessary, as below you show the whole example. It's simple enough below

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,189 @@
...
+      wishlists {
+        id
+        items_count
+      }
+    }
+  }
+`;
+```
+
+When you want to use the load action in component you can do something like this:
```
sethidden, line 170:
```suggestion
When you want to use the load action from the example above in a component you can do something like this:
```

sethidden (changes requested):

docs/guide/global-state-management.md
Frodigo: Actually, I wanted to rewrite graphql plugin to use api.customQuery so it will be good even if we remove graphql-request.

docs/guide/global-state-management.md
Frodigo: ok, changed.

docs/guide/global-state-management.md
Frodigo: ok

docs/guide/global-state-management.md
Frodigo: I would keep this simple.

docs/guide/global-state-management.md
Frodigo: fixed.

docs/guide/global-state-management.md
Frodigo: ok, removed.

docs/guide/global-state-management.md
```diff
@@ -0,0 +1,171 @@
...
+
+In case you want to return only **specific fields**, you need to wrap the field with `computed()`. If you don’t do that, reactivity will break.
+
+```javascript
+import { useWishlistStore } from '~/stores/wishlistStore'
+
+export default {
+  setup() {
+    const wishlistStore = useWishlistStore();
+    const wishlist = computed(() => wishlistStore.wishlist);
```
bartoszherba, line 79: Please, check this https://pinia.vuejs.org/core-concepts/#using-the-store
There is a better pattern to achieve that I think.

bartoszherba (dismissed):
One comment to align more with the solution proposed by Pinia

docs/guide/global-state-management.md
Frodigo: thanks, I changed the code example.

bartoszherba (dismissed):

sethidden (approved):

PR #1075, opened 2022-05-30
fix: example initial fix for search bar not working

ghost:
This PR is not meant to be merged. It's more to show a mimimum brute-force fix and outline why it's not enough. I didn't assign the task to myself in JIRA either - I just came across this while refactoring something else.

### DIFF
--- a/packages/theme/components/Header/SearchBar/SearchResults.vue
+++ b/packages/theme/components/Header/SearchBar/SearchResults.vue
@@ -55,16 +55,7 @@
                   }"
                   :alt="productGetters.getName(product)"
                   :title="productGetters.getName(product)"
-                  :link="
-                    localePath(
-                      `/p/${productGetters.getProductSku(
-                        product
-                      )}${productGetters.getSlug(
-                        product,
-                        product.categories[0]
-                      )}`
-                    )
-                  "
+                  :link="localePath(`/p/${productGetters.getProductSku(product)}`) "
                   :wishlist-icon="isAuthenticated ? 'heart' : ''"
                   :is-in-wishlist-icon="isAuthenticated ? 'heart_fill' : ''"
                   :is-in-wishlist="product.isInWishlist"
@@ -94,13 +85,7 @@
                 }"
                 :alt="productGetters.getName(product)"
                 :title="productGetters.getName(product)"
-                :link="
-                  localePath(
-                    `/p/${productGetters.getProductSku(
-                      product
-                    )}${productGetters.getSlug(product, product.categories[0])}`
-                  )
-                "
+                :link="localePath(`/p/${productGetters.getProductSku(product)}`)"
                 :wishlist-icon="isAuthenticated ? 'heart' : ''"
                 :is-in-wishlist-icon="isAuthenticated ? 'heart_fill' : ''"
                 :is-in-wishlist="product.isInWishlist"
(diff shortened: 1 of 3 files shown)


packages/theme/components/Header/SearchBar/SearchResults.vue
```diff
@@ -55,16 +55,7 @@
...
-                    localePath(
-                      `/p/${productGetters.getProductSku(
-                        product
-                      )}${productGetters.getSlug(
-                        product,
-                        product.categories[0]
-                      )}`
-                    )
-                  "
+                  :link="localePath(`/p/${productGetters.getProductSku(product)}`) "
```
sethidden, line 58: `category` is an optional argument in getProductSku. Since the query does not return "categories" for the product, this caused a "categories is not defined error".

Here I just removed the 2nd argument, because as I said - it's optional, but perhaps it's smarter to add it to the query instead.

packages/theme/modules/catalog/category/composables/useFacet/index.ts
```diff
@@ -35,12 +35,14 @@ export function useFacet(): UseFacetInterface {
...
 
+      const filter = createProductAttributeFilterInput(params);
+
       const productSearchParams: GetProductSearchParams = {
         pageSize,
         search: params.term ? params.term : '',
-        filter: createProductAttributeFilterInput(params),
         sort: createProductAttributeSortInput(params.sort || ''),
         currentPage: params.page,
+        ...(Object.keys(filter).length === 0 ? { filter } : {}),
```
sethidden, line 45: This is a fix for the second cause of the issue. Even if you make createProductAttributeFilterInput return "{}" (ie. "don't include params.categoryId when it's null"), it'll still send the {} in the request body.
That aswell makes the search query return no items. So the above code makes sure filters are not sent if they are an empty object.


There's something potentially wrong with the hard coupling of useFacet to `createProductAttributeFilterInput`, because useFacet's `search` method takes the "filters" param, which is ignored in the current implementation

packages/theme/modules/catalog/category/composables/useFacet/input/createProductAttributeFilterInput.ts
```diff
@@ -8,9 +8,9 @@ export function createProductAttributeFilterInput(params: ComposableFunctionArgs
   const attributeFilter : Record<string, { from: number, to: number } | { eq: unknown } | { in: unknown }> = {};
   const inputFilters = params?.filters ?? {};
 
-  const categoryFilter = {
-    category_id: { in: [params.categoryId, ...inputFilters.category_id ?? []] },
-  };
+  const maybeCategoryFilter = params.categoryId
```
sethidden, line 11: This is a fix for the first cause of the issue. Right now if you try to search in the search box, the variables sent with the query will be `variables={"pageSize":12,"search":"qqq","filter":{"category_id":{"in":[null]}},"sort":{}}`
As you can see, category_id "in": [null] will never find a category, so the search returns 0 products.

sethidden: To me, this requires a bigger refactor because categoryFilter is a subset of attributeFilter - not sure why categoryFilter included as a separate thing. Perhaps useFacet across the code base uses categoryId so much that it's just more comfortable.

Still, I think it'd be - if possible - better to have something like `useFacet.search( { filters: createCategoryFilter(someCategoryId) }` where needed rather than hardcoding it in useFacet

packages/theme/modules/catalog/category/composables/useFacet/index.ts
sethidden: To me this requires a bigger refactor since `useFacet.search`'s params actually have a `filters` field - it's just ignored in the current implementation.

So eithert do `{ ...createProductAAttributeFilterInput(...), ...params.filters }` or remove `createProductAttributeFilterInput` entirely from useFacet and have the user call that manually.

PR #1078, opened 2022-05-30
refactor: refactored useUrlResolver to use the route query

Frodigo:
## Description
THe urlResolver Magento query is deprecated on Magento API side. This PR:
- adds a new route query to the VSF API Client
- updates the useUrlResolver composable to use the new route query instead of depracated urlResolver query.

## Motivation and Context
Refactorization

## Types of changes

## Checklist:

### DIFF
--- a/packages/api-client/src/api/route/index.ts
+++ b/packages/api-client/src/api/route/index.ts
@@ -15,11 +15,11 @@ export type RouteQuery<ROUTE_TYPE> = {
  * @param url the URL to be resolved
  * @param [customQuery] (optional) - custom GraphQL query that extends the default one
  */
-export default async function route<ROUTE_TYPE extends RoutableInterface>(
+export default async function route(
   context: Context,
   url: string,
   customQuery: CustomQuery = { route: 'route' },
-): Promise<ApolloQueryResult<RouteQuery<ROUTE_TYPE>>> {
+): Promise<ApolloQueryResult<RouteQuery<RoutableInterface>>> {
   const { route: routeGQL } = context.extendQuery(customQuery, {
     route: {
       query: routeQuery,
@@ -27,7 +27,7 @@ export default async function route<ROUTE_TYPE extends RoutableInterface>(
     },
   });
 
-  return context.client.query<RouteQuery<ROUTE_TYPE>, QueryRouteArgs>({
+  return context.client.query<RouteQuery<RoutableInterface>, QueryRouteArgs>({
     query: routeGQL.query,
     variables: routeGQL.variables,
   });
--- a/packages/api-client/src/api/route/index.ts
+++ b/packages/api-client/src/api/route/index.ts
@@ -1,0 +1,34 @@
+import type { ApolloQueryResult } from '@apollo/client/core';
+import type { CustomQuery } from '@vue-storefront/core';
+import type { QueryRouteArgs, RoutableInterface } from '../../types/GraphQL';
+import routeQuery from './route';
+import type { Context } from '../../types/context';
+
+export type RouteQuery<ROUTE_TYPE> = {
+  route: ROUTE_TYPE
+};
+
+/**
+ * Returns the canonical URL for a specified product, category, or CMS page
+ *
+ * @param context VSF Context
+ * @param url the URL to be resolved
+ * @param [customQuery] (optional) - custom GraphQL query that extends the default one
+ */
+export default async function route<ROUTE_TYPE extends RoutableInterface>(
+  context: Context,
+  url: string,
+  customQuery: CustomQuery = { route: 'route' },
+): Promise<ApolloQueryResult<RouteQuery<ROUTE_TYPE>>> {
+  const { route: routeGQL } = context.extendQuery(customQuery, {
+    route: {
+      query: routeQuery,
+      variables: { url },
+    },
+  });
+
+  return context.client.query<RouteQuery<ROUTE_TYPE>, QueryRouteArgs>({
+    query: routeGQL.query,
+    variables: routeGQL.variables,
+  });
+}
--- a/packages/api-client/src/api/urlResolver/index.ts
+++ b/packages/api-client/src/api/urlResolver/index.ts
@@ -10,6 +10,7 @@ import type { Context } from '../../types/context';
  * @param context VSF Context
  * @param url the URL to be resolved
  * @param [customQuery] (optional) - custom GraphQL query that extends the default one
+ * @deprecated - use route instead
  */
 export default async function urlResolver(
   context: Context,
--- a/packages/theme/composables/useUrlResolver/index.ts
+++ b/packages/theme/composables/useUrlResolver/index.ts
@@ -4,7 +4,7 @@ import {
   useRoute,
   useContext,
 } from '@nuxtjs/composition-api';
-import type { RoutableInterface } from '@vue-storefront/magento-api';
+import { RoutableInterface } from '~/modules/GraphQL/types';
 import { Logger } from '~/helpers/logger';
 import type { UseUrlResolverErrors, UseUrlResolverInterface } from '~/composables';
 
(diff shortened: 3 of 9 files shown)


packages/api-client/src/types/GraphQL.ts
```diff
@@ -7289,6 +7289,10 @@ export type UrlResolverQueryVariables = Exact<{
 
 export type UrlResolverQuery = { urlResolver?: { id?: number | null | undefined, redirectCode?: number | null | undefined, relative_url?: string | null | undefined, type?: UrlRewriteEntityTypeEnum | null | undefined, entity_uid?: string | null | undefined } | null | undefined };
 
+export type RouteQuery<ROUTE_TYPE> = {
```
sethidden, line 7292: This file is autogenerated. Don't modify it. Or does it come from Magento's GraphQL endpoint? If this is hand-written by you please move it to a separate .ts file

packages/theme/modules/catalog/pages/category.vue
```diff
@@ -190,13 +190,13 @@ export default defineComponent({
 
     const { activeCategory } = useTraverseCategory();
     const activeCategoryName = computed(() => activeCategory.value?.name ?? '');
-    const routeData = ref<EntityUrl>({});
+    const routeData = ref<CategoryTree>(null);
```
sethidden, line 193:
```suggestion
    const routeData = ref<CategoryTree | null>(null);
```

we don't have strict null checks in typescript but pedantism

packages/theme/composables/useUrlResolver/index.ts
```diff
@@ -24,15 +24,15 @@ export function useUrlResolver(): UseUrlResolverInterface {
     search: null,
   });
 
-  const search = async (): Promise<EntityUrl | {}> => {
+  const search = async (): Promise<ROUTE_TYPE> => {
     loading.value = true;
-    let results: EntityUrl = {};
+    let results = null;
```
sethidden, line 29:
```suggestion
    let results : EntityUrl | null = null;
```
Don't remove the type assertion. Right now the type of results is "any"

sethidden: If EntityUrl doesn't work now please use something else that fits

packages/api-client/src/api/urlResolver/index.ts
```diff
@@ -10,6 +10,7 @@ import type { Context } from '../../types/context';
  * @param context VSF Context
  * @param url the URL to be resolved
  * @param [customQuery] (optional) - custom GraphQL query that extends the default one
+ * @deprecated - use route query instead
```
sethidden, line 13: https://github.com/vuestorefront/magento2/pull/1078/files#r884967163
Regarding the linked comment, I think it'd be easier to actually point to what the user is supposed to use. If I read this comment i wouldn't know what to use instead.

whereas
```suggestion
 * @deprecated - use getRoute instead
```
is more exact

sethidden (changes requested):
Left some comments

packages/api-client/src/types/GraphQL.ts
Frodigo: ok

packages/theme/modules/catalog/pages/category.vue
Frodigo: fixed.

packages/theme/composables/useUrlResolver/index.ts
Frodigo: fixed, but I had a problem with wrong generic type and I needed to add @ts-ignore in one line below.

packages/api-client/src/api/route/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { Context } from '../../types/context';
+
+/**
+ * Returns the canonical URL for a specified product, category, or CMS page
+ *
+ * @param context VSF Context
+ * @param url the URL to be resolved
+ * @param [customQuery] (optional) - custom GraphQL query that extends the default one
+ */
+export default async function route<ROUTE_TYPE extends RoutableInterface>(
```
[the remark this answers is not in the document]
Frodigo: I See, but better will be resolveRoute.

packages/api-client/src/api/urlResolver/index.ts
Frodigo: ok, thanks!

packages/api-client/src/api/route/index.ts
```diff
@@ -0,0 +1,34 @@
+import type { ApolloQueryResult } from '@apollo/client/core';
+import type { CustomQuery } from '@vue-storefront/core';
+import type { QueryRouteArgs, RoutableInterface } from '../../types/GraphQL';
+import routeQuery from './route';
+import type { Context } from '../../types/context';
+
+export type RouteQuery<ROUTE_TYPE> = {
```
bartoszherba, line 7: Generic type `ROUTE_TYPE` is redundant and gives nothing. In your code, you do not use any field of method that belongs to a possible generic that actually extends RoutableInteface. All you need is the RoutableInterface because all calls are done on this interface's API. There is no conditional logic for any other type anywhere. You should get rid of the generic type and replace it with RoutableInterface only.

packages/api-client/src/api/route/index.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+/**
+ * Returns the canonical URL for a specified product, category, or CMS page
+ *
+ * @param context VSF Context
+ * @param url the URL to be resolved
+ * @param [customQuery] (optional) - custom GraphQL query that extends the default one
+ */
+export default async function resolveRoute<ROUTE_TYPE extends RoutableInterface>(
```
bartoszherba, line 18: To follow the convention we already have in the API, the method should be named the same as the endpoint route, in this case, `route`. Please, check other index files and methods names. For product the name is the product, for placeOrder the name is placeOrder, and so on.

bartoszherba (changes requested):
I think that the Router type must be removed.

sethidden (changes requested):
Can we please fix the ts-ignore? The error should be fixable, for example with Bartek's suggestions. If not, we can hop on a call

packages/api-client/src/api/route/index.ts
sethidden: https://github.com/vuestorefront/magento2/pull/1078#discussion_r884967163

packages/theme/composables/useUrlResolver/index.ts
sethidden: Wouldn't count ts-ignore as a fix unless there's some third party TS mistake :P Hopefully Bart's comment about removing the generic should help

packages/api-client/src/api/route/index.ts
Frodigo: I don't agree with you. Route query returns a specific type like CategoryTree or SimpleProduct. In case when you need to use a field of that type in the code like that: 

```
const { search: resolveUrl } = useUrlResolver();

routeData.value = await resolveUrl(); 
getContentData(routeData.value?.uid),
```
TS reports a problem because uid does not exist on RoutableInterface type. Then, we need a Generic type here that extends from RoutableInterface.

packages/api-client/src/api/route/index.ts
Frodigo: I see your point and I'm going to restore my change and name the function as `route`

packages/theme/modules/catalog/pages/category.vue
```diff
@@ -111,6 +111,7 @@ import {
   defineComponent, onMounted, ref, ssrRef, useFetch,
 } from '@nuxtjs/composition-api';
 import { CacheTagPrefix, useCache } from '@vue-storefront/cache';
+import type { CategoryTree } from '@vue-storefront/magento-api';
```
bartoszherba, line 114:
```suggestion
import type { CategoryTree } from '~/modules/GraphQL/types';
```

packages/api-client/src/api/route/index.ts
bartoszherba: You are right.

packages/theme/composables/useUrlResolver/index.ts
Frodigo: After chatting with Bartek, the generic type need to stay here. I tried to fix ts-ignore but without success. Could you check out my branch and check that by yourself?

sethidden (dismissed):

bartoszherba (approved):

PR #1079, opened 2022-05-30
refactor: add types for HeaderNavigation*.vue components

ghost:
Add ref-related types to HeaderNavigation. The ref([]) was changed to ref() because of the Vue typedefs. This should be fine as template refs are set during onBeforeMount

### DIFF
--- a/packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
+++ b/packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
@@ -55,11 +55,11 @@
 import {
   SfLink, SfList,
 } from '@storefront-ui/vue';
-import {
-  defineComponent, PropType, ref, onMounted,
-} from '@nuxtjs/composition-api';
-import { CategoryTree } from '~/modules/GraphQL/types';
+import { defineComponent, ref, onMounted } from '@nuxtjs/composition-api';
+import type { PropType } from '@nuxtjs/composition-api';
+import type { CategoryTree } from '~/modules/GraphQL/types';
 import { useUiHelpers } from '~/composables';
+import type { ComponentTemplateRef } from '~/types/componentTemplateRef';
 
 export default defineComponent({
   name: 'HeaderNavigationSubcategories',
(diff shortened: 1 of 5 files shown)


packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
```diff
@@ -79,17 +79,19 @@ export default defineComponent({
...
 
-    const getGroupedLvl2CatRefs = () => {
+    const getGroupedLvl2CatRefs = () : CategoryRef[][] => {
       let current = 0;
-      const result = [];
+      const result : CategoryRef[][] = [];
       lvl1CatRefs.value.forEach((lvl1CatRef) => {
-        const groupCount = Number(lvl1CatRef.$attrs['data-children']);
-        result.push(lvl2CatRefs.value.slice(current, current + groupCount));
+        const groupCount = Number(lvl1CatRef.$el.dataset.children);
```
sethidden, line 92: Changed to start using dataset instead of attrs

sethidden, line 92: Changed to start using dataset instead of attrs

packages/theme/components/Header/Navigation/types.ts
```diff
@@ -0,0 +1,9 @@
+import type { ComponentInstance } from '@nuxtjs/composition-api';
+
+/**
+ * In HeaderNavigation*.vue components, template refs are applied to components, not HTML elements,
+ * so this type uses Vue's ComponentInstance interface rather than HTMLElement.
+ *
+ * ComponentInstance["$el"] is `Element` by default, which doesn't have `.focus()``
+ */
+export type CategoryRef = ComponentInstance & { $el: HTMLElement };
```
bartoszherba, line 9: Can we replace the `CategoryRef` with a more generic name? If you hide the name the type definition is much more universal than just in category context

bartoszherba (dismissed):

packages/theme/components/Header/Navigation/types.ts
sethidden: done

packages/theme/components/Header/Navigation/HeaderNavigationSubcategories.vue
Frodigo: What is the advantage of this change?

sethidden: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

whenever you put `<div data-something="abcd"/>` the data attribute gets put into a special "dataset" property on the DOM element.

myDomElement.dataset.something

I felt it's weird that the data is pulled from the Vue attribute, because infact this is a regulard HTML attribute on the element

RetroModernDev (approved):

bartoszherba (approved):

PR #1080, opened 2022-05-31
refactor: improve typing of sorting in facetGetters/category.vue

sethidden:
Fix issue where AgnosticSort was totally different type than what would appear at runtime

### DIFF
--- a/packages/theme/modules/catalog/category/getters/facetGetters.ts
+++ b/packages/theme/modules/catalog/category/getters/facetGetters.ts
@@ -1,65 +1,50 @@
-import type {
-  AgnosticPagination,
-  AgnosticSort,
-} from '~/composables/types';
+import { perPageOptions } from '~/modules/catalog/category/composables/useFacet/perPageOptions';
+import type { ProductInterface } from '~/modules/GraphQL/types';
+import type { AgnosticPagination } from '~/composables/types';
+import type { SortingModel } from '~/modules/catalog/category/composables/useFacet/sortingOptions';
+import type { UseFacetSearchResult } from '~/modules/catalog/category/composables/useFacet/useFacet';
 
-import { ProductInterface } from '~/modules/GraphQL/types';
-import { FacetSearchParams } from '~/modules/catalog/category/composables/useFacet';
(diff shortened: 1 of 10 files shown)


packages/theme/modules/catalog/category/getters/facetGetters.ts
```diff
@@ -1,65 +1,50 @@
...
-  data: S;
-  input: FacetSearchParams;
+import { perPageOptions } from '~/modules/catalog/category/composables/useFacet/perPageOptions';
+import type { ProductInterface } from '~/modules/GraphQL/types';
+import type { AgnosticPagination } from '~/composables/types';
+import type { SortingModel } from '~/modules/catalog/category/composables/useFacet/sortingOptions';
+import type { UseFacetSearchResult } from '~/modules/catalog/category/composables/useFacet/useFacet';
+
+export interface FacetsGetters {
+  getSortOptions: (searchData: UseFacetSearchResult) => SortingModel;
```
sethidden, line 8: Previously the type here was `AgnosticSort` which turned out to be completely different at runtime. TypeScript didn't report an error since the searchData argument was any.

Frodigo (approved):

packages/theme/modules/catalog/category/getters/facetGetters.ts
```diff
@@ -1,65 +1,50 @@
...
 
-const getSortOptions = (searchData): AgnosticSort => {
-  if (!searchData || !searchData.data || !searchData.data.availableSortingOptions) {
     return {
-      options: [],
-      selected: '',
-    } as AgnosticSort;
-  }
+      options: searchData.data.availableSortingOptions,
+      selected: searchData.input.sort,
```
bartoszherba, line 24: What if there will be no sort param?

bartoszherba (approved):

packages/theme/modules/catalog/category/getters/facetGetters.ts
sethidden: CategoryNavbar will not show any sort option. This will only happen if someone puts a sort id that doesn't exist into the url.

PR #1095, opened 2022-06-03
refactor!: break down login modal into separate components

ghost:
Before this PR, LoginModal was a big component which contained 3 forms in a single template.

After this PR, there are 3 separate components for forms, and one parent component for all 3 forms which handles the api queries etc.

Added tests aswell.


packages/theme/components/LoginModal/__tests__/LoginFragment.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+import userEvent from '@testing-library/user-event';
+import { createLocalVue } from '@vue/test-utils';
+import { PiniaVuePlugin } from 'pinia';
+import { createTestingPinia } from '@pinia/testing';
+
+import { extend } from 'vee-validate';
+import { required, email } from 'vee-validate/dist/rules';
+import LoginFragment from '../LoginFragment.vue';
+
+extend('required', { ...required, message: 'error:required' });
```
sethidden, line 11: This is not ideal as it makes the test dependent on implementation

packages/theme/components/LoginModal/LoginModal.vue
```diff
@@ -0,0 +1,244 @@
+<template>
```
sethidden, line 1: remember to move all of those to the customer module

packages/theme/components/LoginModal/__tests__/LoginFragment.spec.ts
sethidden: gotta browse https://vee-validate.logaretm.com/v4/guide/testing more to find a better solution

packages/theme/components/LoginModal/RegisterFragment.vue
```diff
@@ -0,0 +1,168 @@
...
+          :label="$t('Sign Up for Newsletter')"
+          name="signupNewsletter"
+          class="form__element"
+        />
+        <ValidationProvider
+          v-slot="{ errors }"
+          :rules="{ required: { allowFalse: false } }"
+        >
+          <SfCheckbox
+            v-model="formCopy.isWillToCreateAccountConfirmed"
```
sethidden, line 82: This shouldn't be part of the emitted payload

packages/theme/components/LoginModal/__tests__/LoginFragment.spec.ts
sethidden: removed the extensions and relying on hardcoded strings for now

packages/theme/components/LoginModal/LoginModal.vue
sethidden: done

packages/theme/components/LoginModal/RegisterFragment.vue
sethidden: I'll keep it for now and remove it in the parent LoginModal

sethidden:
Need to check if all the :loading props and the error slots work fine

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
```diff
@@ -0,0 +1,251 @@
...
+      }
+    };
+
+    const onForgotPasswordFormSubmit = async (form: ForgotPasswordFormFields) => {
+      forgotPasswordForm.value = form;
+      const { token, cleanup } = await getRecaptchaInfo(isRecaptchaEnabled.value);
+      await resetPassword({ email: form.username, ...(token ? { recaptchaToken: token } : {}) });
+      cleanup();
+      currentlyDisplayedForm.value = 'forgotPasswordThankYou';
+    };
```
sethidden, line 218: There is a bit of duplication here but I prefer duplication to making some weird abstraction/function.

There's an abstraction now in LoginModal on develop and it just doesn't work - it's used for login and register, but a separate function exists for forgotpassword as the abstraction doesn't fit.

packages/theme/modules/customer/components/LoginModal/__tests__/LoginModal.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+      localVue,
+      pinia: createTestingPinia(),
+    });
+    const loginForm = getByTestId('login-form');
+    const form : LoginFormFields = {
+      email: 'james@bond.co.uk',
+      password: 'P@ssw0rd!',
+    };
+    await waitFor(() => {
+      (loginForm as HTMLElementWithVue).__vue__.$emit('submit', form);
```
sethidden, line 58: Note I'm accessing the Vue instance which is authors of testing-library/vue don't intend me to do. The reason I'm doing it is that I've already tested the form behavior in LoginForm.spec.ts and I know it emits submit correctly.

Basically, I wanted it to be more of a unit test rather than an integration test.

packages/theme/modules/customer/components/LoginModal/forms/__tests__/ForgotPasswordForm.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { render } from '@testing-library/vue';
+import userEvent from '@testing-library/user-event';
+
+import ForgotPasswordFragment from '../ForgotPasswordForm.vue';
+import { ForgotPasswordFormFields } from '../types';
+
+describe('ForgotPasswordForm', () => {
```
sethidden, line 7: It's easy to test forms thanks to the props approach.

packages/theme/modules/customer/components/LoginModal/forms/__tests__/RegisterForm.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { render } from '@testing-library/vue';
+import userEvent from '@testing-library/user-event';
+
+import RegisterForm from '../RegisterForm.vue';
+import { RegisterFormFields } from '../types';
+
+describe('RegisterForm', () => {
+  it('emits submit if validation passes', async () => {
```
sethidden, line 8: TODO: add more register form tests

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
```diff
@@ -0,0 +1,251 @@
...
+import { useUser } from '~/modules/customer/composables/useUser';
+import { customerPasswordRegExp, invalidPasswordMsg } from '~/modules/customer/helpers/passwordValidation';
+
+import LoginForm from './forms/LoginForm.vue';
+import RegisterForm from './forms/RegisterForm.vue';
+import ForgotPasswordForm from './forms/ForgotPasswordForm.vue';
+import ForgotPasswordThankYou from './forms/ForgotPasswordThankYou.vue';
+import { ForgotPasswordFormFields, LoginFormFields, RegisterFormFields } from './forms/types';
+
+extend('email', {
```
bartoszherba, line 99: Do we need these extensions here? We have them in forms, and in the LoginModal there is no form nor validator.

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
```diff
@@ -0,0 +1,251 @@
...
+  useContext,
+} from '@nuxtjs/composition-api';
+import {
+  SfModal,
+  SfBar,
+} from '@storefront-ui/vue';
+import { extend } from 'vee-validate';
+import { required, email } from 'vee-validate/dist/rules';
+import { useUiState } from '~/composables/useUiState';
+import useCart from '~/modules/checkout/composables/useCart';
```
bartoszherba, line 87: Please, once you use  { } once you do not. Unify this. (compare useCart and useUser import)

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
```diff
@@ -0,0 +1,251 @@
...
+  components: {
+    SfModal,
+    SfBar,
+
+    LoginForm,
+    RegisterForm,
+    ForgotPasswordForm,
+    ForgotPasswordThankYou,
+  },
+  setup(_props, { emit }) {
```
bartoszherba, line 127: why props are underscored? if you are not using props use just `_`
```suggestion
  setup(_, { emit }) {
```

packages/theme/modules/customer/components/LoginModal/forms/ForgotPasswordThankYou.vue
```diff
@@ -0,0 +1,25 @@
...
+  <div>
+    <i18n
+      tag="p"
+      path="forgotPasswordConfirmation"
+    >
+      <span class="bold">
+        <slot name="email" />
+      </span>
+    </i18n>
+    <p>{{ $t('Thank You Inbox') }}</p>
```
bartoszherba, line 11: What is this? ;_) What does it mean?

packages/theme/modules/customer/components/LoginModal/forms/ForgotPasswordThankYou.vue
```diff
@@ -0,0 +1,25 @@
...
+  </div>
+</template>
+
+<script lang="ts">
```
bartoszherba, line 15: Is this required?
```
<script lang="ts">
import { defineComponent } from '@nuxtjs/composition-api';

export default defineComponent({});
</script>
```

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
```diff
@@ -0,0 +1,251 @@
...
+          <template #error>
+            {{ userError.login && userError.login.message }}
+          </template>
+        </LoginForm>
+        <RegisterForm
+          v-else-if="currentlyDisplayedForm === 'register'"
+          data-testid="register-form"
+          :show-recaptcha="isRecaptchaEnabled"
+          :loading="loading"
+          :form="registerForm"
```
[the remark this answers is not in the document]
bartoszherba: Ok, it is easier to test form but I still do not understand what do we have formCopy and watcher.

bartoszherba (changes requested):

packages/theme/modules/customer/components/LoginModal/forms/ForgotPasswordThankYou.vue
sethidden: I moved it over from the old LoginModal, I did not change the labels, unless by mistake

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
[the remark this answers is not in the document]
sethidden: In a situatioin where fromCopy doesn't exist, I'd be forced to something like this:

```
props: { form: { type: Object as PropType<...> }

<input v-model="form.lastName"
```

this obviously equivalent to `<input :value="form.lastName" @input="form.lastName = $event`

So as you can see, this would be modifying props directly, which is not only illegal, but throws erros in the console

This is also necessary because I need some variable to buffer the changes made by the user
1. inital form is passed to LoginForm
2. user makes some changes in the form, but those changes exist only in formCopy
3. once the form validation lets me know that what the user typed in is fine, only then I emit formCopy, which replaces the real, non-copy loginForm ref

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
```diff
@@ -0,0 +1,251 @@
...
+    >
+      <div>
+        <LoginForm
+          v-if="currentlyDisplayedForm === 'login'"
+          data-testid="login-form"
+          :show-recaptcha="isRecaptchaEnabled"
+          :loading="loading"
+          :form="loginForm"
+          @submit="form => onLoginFormSubmit(form)"
+          @go-to-register="currentlyDisplayedForm = 'register'"
```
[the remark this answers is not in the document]
sethidden: I thought about this while writing this code. But I decided that I'll keep it as-is because I don't want the Login**Form** to know anything about Login**Modal** logic (currently displayed form). LoginForm is just a form - has some inputs then emits an object - there's no reason for it to hold "currentlyDisplayedForm" logic. Clicking the "go to register" button emits the go to register event

What the parent component does with that event is up to it and the logic it handles

As for extracting to function, there's too little logic here to warrant extracting. I prefer the at-a-glance assignment in the template since you don't have to look at the method to see what it does.

I've thout about maybe changing go-to-register to navigate:register or something

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
```diff
@@ -0,0 +1,251 @@
...
+        @click:close="$emit('close')"
+      />
+    </template>
+    <transition
+      name="sf-fade"
+      mode="out-in"
+    >
+      <div>
+        <LoginForm
+          v-if="currentlyDisplayedForm === 'login'"
```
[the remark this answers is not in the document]
sethidden: As in previous comment

>As for extracting to function, there's too little logic here to warrant extracting. I prefer the at-a-glance assignment in the template since you don't have to look at the method to see what it does.

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
sethidden: Right. Removed in f902058080812c5dd614f223cab5193a4474be95

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
sethidden: I don't think this is important as there's no difference, but I changed it in c3e6870daaea409e8d30683a721b59feaf8e103f because it's quicker than arguing

packages/theme/modules/customer/components/LoginModal/forms/ForgotPasswordThankYou.vue
sethidden: I think the original author meant "inbox message (thanks, check your inbox) for the forgot password thank you page"

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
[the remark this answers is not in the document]
bartoszherba: Than is why functions can be named ;)

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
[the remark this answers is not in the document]
bartoszherba: Now  LoginForm knows about the registration form and even more, it expects the register form and even makes sense if and only when RegisterForm exists. On the other hand, a generic go-to event assumes nothing about anything and can be eventually handled by ANY parent in ANY way. go-to-register is more like a well-defined command than an event and go-to event emitter knows nothing about any other component as it can be handled anywhere not just in LoginModal

packages/theme/modules/customer/components/LoginModal/forms/ForgotPasswordThankYou.vue
bartoszherba: No, it is not. Neither in vscode or webstorm. But what is highlighted in both is (regardless of the defineComponent).

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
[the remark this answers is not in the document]
Frodigo: I agree with Bartosz here, once we use events like `@go-to-register`, `@go-to-forgot-password`, etc, we need to add a new event when we want to add a new form (or component) inside the component. When we use the `@go-to` event with an argument, it makes the LoginModal component more generic and readable.

packages/theme/modules/customer/components/LoginModal/forms/ForgotPasswordThankYou.vue
sethidden: image

whereas it's fine with defineComponent

image

I assume you're using Vetur, not Volar? (webstorm has its own inhouse extension).

Having the script isn't a big deal. I'll resolve this.

packages/theme/modules/customer/components/LoginModal/__tests__/LoginModal.spec.ts
Frodigo: It's not about being more of a unit test rather than an integration test. It's about showing/ not showing implementation details in tests.

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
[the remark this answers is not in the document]
sethidden: fa43904764991d4d5204dde500fb0fbc11ae80c5

packages/theme/modules/customer/components/LoginModal/LoginModal.vue
bartoszherba: Probably, but this is a common approach in many docs, blogs, etc.

RetroModernDev (approved):

bartoszherba (approved):

PR #1100, opened 2022-06-07
fix: removed filters skeleton on mobile devices

Frodigo:
### Before 
before this change on the category page, there were filters loading skeleton on mobile devices, but on mobile devices, filters can be visible only in overlay

### After
After this change loader skeleton for filters on the category page is not visible anymore on mobile devices.


bartoszherba (approved):
After a deep investigation I believe that this is a good change.

PR #1103, opened 2022-06-08
test: add CategorySidebar tests

ghost:
Bring old, skipped CategorySidebar tests back to life and add some more tests that were asked for in M2-631


packages/theme/components/__tests__/CartSidebar.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+      );
+
+      const secondProductElement = getAllByTestId('cart-sidebar-collected-product')[1];
+      const increaseQuantityButton = within(secondProductElement).getByTestId('increase');
+      userEvent.click(increaseQuantityButton);
+
+      await waitFor(() => {
+        const { uid: uidOfSecondProduct } = useCartMockInstance.cart.value.items[1];
+        expect(useCartMockInstance.updateItemQty).toHaveBeenCalledWith({
+          product: expect.objectContaining({ uid: uidOfSecondProduct }),
```
sethidden, line 168: Note that I'm not using "soft" matching for the product object. The product object passed here is huge, but due to a computed inside CartSidebar, the items returned by useCart are modified a bit, so `{ product }` passed here is a little different than `useCartMockInstance.cart.value.items[1]`

`updateItemQty`'s implementation relies only on the `product.uid` so this is mostly a fault of the composable fn requiring tons of properties it doesn't need

sethidden: I made a task for this: https://vsf.atlassian.net/browse/M2-810

RetroModernDev (approved):

PR #1109, opened 2022-06-09
chore: use lodash-es over per-function packages

ghost:
lodash per-function packages are discouraged in the lodash docs (see M2-720 description for more details).

eg. lodash.debounce was last updated 6 years ago, yet lodash-es's repo has changed to debounce 3 years ago, so the per-package modules are not updated

In case you're worried this will increase the bundle size since there's one big package rather than 3 small ones - the difference between the `lodash-es` and `lodash` package on NPM is that `lodash-es` is published as an ES module, so doing things like

```
import _ from 'lodash-es'
_.debounce(() => {})
```

will put only files related to debounce into the bundle, not whole lodash (if you wrote `from 'lodash'` though, it'd import it all)

The main bundle generated in this PR is 500 bytes heavier than develop, but avoids the pitfalls mentioned in M2-720 lodash docs link


yarn.lock
```diff
@@ -18725,13 +18736,13 @@ terser@^4.1.2, terser@^4.6.13, terser@^4.6.3:
     source-map-support "~0.5.12"
 
 terser@^5.3.4, terser@^5.7.2:
-  version "5.11.0"
-  resolved "https://registry.yarnpkg.com/terser/-/terser-5.11.0.tgz#2da5506c02e12cd8799947f30ce9c5b760be000f"
-  integrity sha512-uCA9DLanzzWSsN1UirKwylhhRz3aKPInlfmpGfw8VN6jHsAtu8HJtIpeeHHK23rxnE/cDc+yvmq5wqkIC6Kn0A==
```
sethidden, line 18730: I also messed with the terser in yarn.lock a bit - hopefully now `yarn add` works fine without the need for the lerna trick. I welcome you to checkout this branch and see for yourself if eg. `yarn add is-odd` (example package name) works

RetroModernDev (approved):

PR #1112, opened 2022-06-10
chore: use lodash-es over per-function packages

ghost:
(same as earlier PR, just added build.transpile)


packages/theme/nuxt.config.js
```diff
@@ -207,6 +207,7 @@ export default () => {
       ],
       transpile: [
         'vee-validate',
+        'lodash-es',
```
sethidden, line 210: This fixes the earlier error

RetroModernDev (approved):

PR #1115, opened 2022-06-10
fix: hide mobile address delete buttons on default addresses

ghost:
this already works fine on desktop, but was not implemented in mobile


sethidden:
wrong task in branch name

PR #1138, opened 2022-06-17
fix: overlapping homepage app store buttons

ghost:
Before:
image
After:
image

### DIFF
--- a/packages/theme/components/MobileStoreBanner.vue
+++ b/packages/theme/components/MobileStoreBanner.vue
@@ -13,13 +13,15 @@
           data-testid="banner-cta-button"
           link="#"
         >
-          <SfImage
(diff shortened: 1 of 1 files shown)


packages/theme/components/MobileStoreBanner.vue
```diff
@@ -13,27 +13,31 @@
...
           link="#"
         >
-          <SfImage
-            src="/homepage/apple.png"
-            loading="lazy"
-            alt="Apple"
-            :width="134"
-            :height="44"
-          />
+          <div class="app-store-image">
```
sethidden, line 16: can't apply class directly to SfImage because it doesn't v-bind attrs

RetroModernDev (approved):

PR #1143, opened 2022-06-20
fix: pre-select default shipping address

ghost:
Before this Pr the default shipping/billing address wouldn't get preselected

image


sethidden:
And well, it doesn't fill the from so it's not functioal

RetroModernDev (dismissed):

RetroModernDev (approved):

PR #1146, opened 2022-06-20
fix: product pricing

Frodigo:
WIP: fixing pricing issues on PDP.


packages/theme/modules/catalog/pages/product.vue
```diff
@@ -126,6 +131,18 @@ export default defineComponent({
...
+    onMounted(async () => {
+      if (!product.value) {
+        await fetchProduct();
+      }
+
+      const { data } = await query<ProductDetailsQuery>(getProductPriceBySkuGql, { sku: product.value.sku });
+
+      product.value = {
+        ...merge(product.value, data.products?.items?.[0] as Product),
+      };
```
sethidden, line 143: https://docs-lodash.com/v4/merge/

merge mutates its first argument. I think you're looking for

```suggestion
      product.value = merge({}, product.value, data.products?.items?.[0] as Product);
```

packages/theme/modules/catalog/product/components/product-types/bundle/BundleProduct.vue
```diff
@@ -89,10 +91,18 @@
           class="product__description desktop-only"
         />
         <BundleProductSelector
+          v-if="productPrice"
           :can-add-to-cart="canAddToCart(product, qty)"
           :product="product"
           @update-price="customizationPrice = $event"
         />
+        <div v-else>
+          <client-only>
```
sethidden, line 100: Why the client-only here? We don't do that in any of the other skeletons

packages/theme/modules/catalog/product/components/product-types/bundle/BundleProductOptionSkeleton.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+  <div>
+    <div>
```
sethidden, line 3: nitpick: are the double nested divs necessary? ie. is 1 not enough?

packages/theme/modules/catalog/product/components/product-types/bundle/BundleProduct.vue
```diff
@@ -195,9 +208,9 @@ export default defineComponent({
       () => props.product?.description?.html || '',
     );
 
-    const productPrice = ref(getProductPrice(props.product).regular);
-    const productSpecialPrice = ref(getProductPrice(props.product).special);
-    const productMaximumPrice = ref(getProductPrice(props.product).maximum);
+    const productPrice = computed(() => getProductPrice(props.product).regular);
```
sethidden, line 211: ignore if you want because this is a "by the way" refactor: why call getProductPrice() three times? Wouldn't it just be possible to `const productPrices = computed(() => getProductPrice(props.product))` then refer to it in the template through `productPrices.regular` etc?

packages/theme/modules/catalog/pages/product.vue
```diff
@@ -126,6 +131,18 @@ export default defineComponent({
       addTags([...tags, ...productTags]);
     });
 
+    onMounted(async () => {
+      if (!product.value) {
+        await fetchProduct();
```
sethidden, line 136: Why is this called both in useFetch and onMounted?

packages/theme/modules/catalog/pages/product.vue
sethidden: (without the spread ... is fine, as it'll be reference-free aswell)

sethidden (changes requested):

packages/theme/modules/catalog/pages/product.vue
Frodigo: Because when you go to PDP direclty from browser for example by clicking product on category page, product value equals null. I tried to use the `ssrRef` instead of the `ref` for the product, but it didn't work. If you have any ideas how to do that better I am open for discussion.

sethidden: Sounds like this can be fixed by putting `v-if="$fetchState.pending` somewhere near the top of the component root or changing some code to handle the fact that product.value can be null - it's useFetch after all, not useAsync

packages/theme/modules/catalog/product/components/product-types/bundle/BundleProduct.vue
Frodigo: Refactored.

packages/theme/modules/catalog/product/components/product-types/bundle/BundleProductOptionSkeleton.vue
Frodigo: My mistake. fixed.

packages/theme/modules/catalog/product/components/product-types/bundle/BundleProduct.vue
Frodigo: I tested something and forgot to remove it. Now it's ok.

packages/theme/modules/catalog/pages/product.vue
sethidden: spread is ok but what about the first argument `{}`? Can it be added?

You should theoretically be able to do
```
product.value = merge({}, product.value, data.products?.items?.[] as Product);
```
or
```
merge(product.value, data.products?.items?.[] as Product)
```

packages/theme/modules/catalog/pricing/getGroupedProductPriceCommand.ts
```diff
@@ -8,8 +8,8 @@ export function getGroupedProductPriceCommand(product: GroupedProduct): number {
     return regular > special && special !== null ? special : regular;
   };
 
-  return product.items.reduce(
+  return product.items ? product.items.reduce(
```
sethidden, line 11:
```suggestion
  return product.items?.reduce(...) ?? 0
```


sry for 2nd round, didn't notice before

packages/theme/modules/catalog/product/components/product-types/bundle/BundleProduct.vue
```diff
@@ -195,10 +206,11 @@ export default defineComponent({
       () => props.product?.description?.html || '',
     );
 
-    const productPrice = ref(getProductPrice(props.product).regular);
-    const productSpecialPrice = ref(getProductPrice(props.product).special);
-    const productMaximumPrice = ref(getProductPrice(props.product).maximum);
-    const customizationPrice = ref(productSpecialPrice.value ?? productPrice.value);
+    // const productPrice = computed(() => getProductPrice(props.product).regular);
```
sethidden, line 209: hmmm

packages/theme/modules/catalog/pages/product.vue
Frodigo: Fixed, thanks!

packages/theme/modules/catalog/pages/product.vue
Frodigo: Fixed.

packages/theme/modules/catalog/pricing/getGroupedProductPriceCommand.ts
Frodigo: Fixed.

packages/theme/modules/catalog/product/components/product-types/bundle/BundleProduct.vue
Frodigo: fixed

sethidden (approved):

PR #1151, opened 2022-06-21
ci: create gh deployment entry as early as possible

ghost:
This PR starts a deployment as soon as the `deploy to vsf cloud` action starts. Previously it only created a deployment after the build - before the curl to vsf cloud / farmer. It's worth mentioning the build->deploy process is a single workflow. We don't upload artifacts in separate steps or anything.

The previous implementation was insufficient as today in the morning there's been some NPM failures which broke the `deploy to vsf cloud` workflow builds, but the deployments did not appear at all. This led to our QA seeing the tasks in JIRA have the QA status, but not being able to tell if they're already deployed or not.

Now if a situation like this happens again, a deployment will exist and it will show up as failed.

As you can see by the number of deployments below, I've checked if this PR actually works by using the defunct enterprise environment for testing.

### DIFF
--- a/.github/workflows/deploy-vue-storefront-cloud.yml
+++ b/.github/workflows/deploy-vue-storefront-cloud.yml
@@ -8,15 +8,62 @@ on:
       - release/*
 
 jobs:
+  create-deployment:
+    runs-on: ubuntu-latest
+    outputs:
+      environment-name: ${{ steps.determine-environment.outputs.name }}
+      environment-code: ${{ steps.determine-environment.outputs.code }}
+      deployment_id: ${{ steps.deployment.outputs.deployment_id }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-vue-storefront-cloud.yml
```diff
@@ -65,54 +112,33 @@ jobs:
 
           LAST_COMMIT: ${{ github.sha  }}
 
-  deploy-main:
```
sethidden, line 68: No need for multiple deploys blocks now. They were only for assigning the correct environment name to a branch, but now this is handled by the bash script at the top of the file.

RetroModernDev (approved):

PR #1154, opened 2022-06-22
fix: show country/region names instead of codes

ghost:
The Jira issue actually asks only to fix the regions, but I also fixed the countries. I asked QA about their thoughts

M2-849


sethidden:
In draft because waiting for QA to respond

sethidden:
image

RetroModernDev (approved):

PR #1155, opened 2022-06-22
ci: make deployment 'success' status more accurate

ghost:
report the deployment as complete only when we're somewhat sure the new
image has been deployed

before this commit the workflow worked like "ask vsf cloud to deploy the
container" and immediately reported the deployment as finished, even
though the actual images are replaced ~5 minutes later

M2-857


RetroModernDev (approved):
Thanks for that!

PR #1159, opened 2022-06-23
fix: cart store not cleared properly on logout

ghost:
Before this commit if you - while logged in - added some new items to
the cart, then logged out, the cart store would still contain the full,
old cart object

This was caused by the useCart().clear() method patching the cart
store (similarly to useUser().setCart()), and it seems like there was a
race condition there. Please note .clear() re-downloads the cart

M2-831

### DIFF
--- a/packages/theme/modules/customer/pages/MyAccount/useSidebarLinkGroups.ts
+++ b/packages/theme/modules/customer/pages/MyAccount/useSidebarLinkGroups.ts
@@ -49,7 +49,8 @@ export const useSidebarLinkGroups = () => {
           label: 'Log out',
           listeners: {
             click: async () => {
-              await Promise.all([logout({}), clear({})]);
+              await logout({});
+              await clear({});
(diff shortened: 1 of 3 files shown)


sethidden:
Should be ok now

packages/theme/modules/customer/pages/MyAccount/useSidebarLinkGroups.ts
```diff
@@ -49,7 +49,8 @@ export const useSidebarLinkGroups = () => {
           label: 'Log out',
           listeners: {
             click: async () => {
-              await Promise.all([logout({}), clear({})]);
+              await logout({});
```
Frodigo, line 52: What is advantage of this change?

Frodigo: ok, sorry I didn't catch that.

RetroModernDev (approved):

PR #1170, opened 2022-06-27
fix: unify login/register notifications and error handling

sethidden:
Before this commit, useUser().register error path used useUiNotification().sendNotification,
but useUser().login's error path did not. UserAccount.vue used sendNotification
aswell on its own side.

So when login() errored (bad password etc.) the useUser().login didn't
send notifications, but UserAccount.vue - further down the pipeline - did.

The contrary happened in register() - both UserAccount.vue and
register() sent notifications.

So I had to make a decision - either add sendNotification to
useUser().login the same way as in register, AND remove the
UserAccount.vue notifications

or

remove notifications from the composable and just handle it in
UserAccount.vue itself.

We decided on option 1 with Marcin.

M2-752

### DIFF
--- a/packages/theme/modules/customer/components/LoginModal/LoginModal.vue
+++ b/packages/theme/modules/customer/components/LoginModal/LoginModal.vue
@@ -26,11 +26,7 @@
           :form="loginForm"
           @submit="form => onLoginFormSubmit(form)"
           @change-form="currentlyDisplayedForm = $event"
-        >
-          <template #error>
-            {{ userError.login && userError.login.message }}
-          </template>
-        </LoginForm>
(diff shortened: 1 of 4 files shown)


packages/theme/modules/customer/components/LoginModal/LoginModal.vue
```diff
@@ -26,11 +26,7 @@
           :form="loginForm"
           @submit="form => onLoginFormSubmit(form)"
           @change-form="currentlyDisplayedForm = $event"
-        >
```
sethidden, line 29: now handled by useUser()'s internal sendNotification

PR #1178, opened 2022-06-29
fix:  fixed wrong values of selected configurable product variants

Frodigo:
Before this PR, the last option of configurable was added to the cart instead of the option selected by the user.

M2-860
Screenshot 2022-06-29 at 08 13 30


sethidden (dismissed):
Left one comment but it's not that important

sethidden (approved):

Frodigo:
@sethidden I added your suggestion, please approve again.

PR #1199, opened 2022-07-01
fix!: m2-668 fixed price recalculation for configurable products

Frodigo:
The second part of https://github.com/vuestorefront/magento2/pull/1146

### Before
Before this PR, prices were not recalculated properly for configurable products.
Reason: prices for product variants were cached in Redis. 

### After
After this PR prices are recalculated properly for configurable products. 

### Solution
I moved loading data of variants from a query that is run on server-side (`packages/api-client/src/api/productDetail/productDetailsQuery.ts`) to the query that is run on client-side (`packages/theme/modules/catalog/product/queries/getProductPriceBySku.gql.ts`)

M2-668

### DIFF
--- a/packages/theme/modules/catalog/product/queries/getProductPriceBySku.gql.ts
+++ b/packages/theme/modules/catalog/product/queries/getProductPriceBySku.gql.ts
@@ -24,11 +24,60 @@ const fragmentPriceRangeFields = `
 `;
 
 export default `
-  query getProductPriceBySku($sku: String) {
-    products(filter: {sku: {eq: $sku}}) {
+  query getProductPriceBySku(
+    $filter: ProductAttributeFilterInput,
+    $configurations: [ID!]
+  ) {
+    products(filter: $filter) {
       items {
         price_range {
           ...PriceRangeFields
+        }
+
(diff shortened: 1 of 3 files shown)


sethidden (approved):

packages/theme/modules/catalog/product/queries/getProductPriceBySku.gql.ts
```diff
@@ -24,13 +24,62 @@ const fragmentPriceRangeFields = `
...
+              }
+              regular_price {
+                currency
+                value
+              }
+            }
+          }
+
+          configurable_product_options_selection(configurableOptionValueUids: $configurations) {
+            options_available_for_selection {
```
bartoszherba, line 62: I am just curious, why do we need all that not price-related stuff in this query?

Frodigo: I met some issues with merging objects and the solution was complicated I decided to simplify it for now. 

If this causes some performance issues, we will re-thing the solution. So let's observe the speed curve tests, and we will see, my friend.

PR #1232, opened 2022-07-08
feat: secure cookie attribute is not set

bartoszherba:
## Description
I have added configurable cookies options so it will be possible to customize cookies set mods.

## Types of changes

## Checklist:

### DIFF
--- a/docs/installation-setup/configure-integration.md
+++ b/docs/installation-setup/configure-integration.md
@@ -71,3 +71,46 @@ When working with translation in your application, you need to:
 
 1. Add translations in Magento for products and categories.
 2. Update the `i18n.locales` array in the `nuxt.config.js` file and add translations to the corresponding files in the `lang` directory.
+
+### 4. Configure default cookies settings
+
+Vue Storefront app uses different cookies but all share the same default config. To adjust the configuration you have to modify `middleware.config.js`.
+Once done, rebuild your application.
+
+```js
+module.exports = {
+  integrations: {
+    magento: {
+      configuration: {
+        /*...*/
+        // Here you can override default cookies options
+        cookiesDefaultOpts: {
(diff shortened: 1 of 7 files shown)


packages/theme/middleware.config.js
```diff
@@ -12,6 +12,10 @@ module.exports = {
         cookies: {
           ...cookieNames,
         },
+        cookiesDefaultOpts: {
+          httpOnly: false,
+          secure: false,
```
bartoszherba, line 17: If I set this === true then we will have to change it for local development if we do not have ssl, which is true in most cases. If I leave it as it is, cookies will remain non-secure on our demo envs.

Frodigo: As we talked about in person, it would be great to enable it in our environments using nuxt.config.addtional.js file.

bartoszherba: 1. I set the flag to true as we should setup the app to be as safe as possible OOTB
2. I added the instruction of how to setup SSL certificate (it is straightforward and will take no more than 1 min)

docs/installation-setup/configure-integration.md
```diff
@@ -71,3 +71,59 @@ When working with translation in your application, you need to:
...
+        /*...*/
+      },
+    },
+  },
+};
+```
+
+### 5. Using HTTPS configuration for a local development
+
+### Install `mkcert`
```
Frodigo, line 100: Can you add information how to install it on Windows? In addition, please add link to tool that you use: https://github.com/FiloSottile/mkcert

Frodigo (changes requested):
Looks good, one request about adding docs for Windows users

Frodigo (approved):

PR #1237, opened 2022-07-11
refactor: added possibility to enable customer email confirmation

Frodigo:
### Before
Before this PR, when the admin set up config: Stores - > configuration -> customesr -> customer confiuguration -> Require Emails Confirmation to yes, the user didn't see the appropriate notification, and the login action of useUser composable returned an error. 

### After
- it's possible to set up the `customer_create_account_confirm` flag in middleware.config.js
- when the flag equals true, then the user can see the notification "You must confirm your
 account. Please check your email for the confirmation link". The register action is not triggered
- when the flag equals false, the user can register immediately after log-in
Screenshot 2022-07-11 at 12 37 26

### Documentation: 
Screenshot 2022-07-12 at 11 33 39
Screenshot 2022-07-12 at 11 33 54


bartoszherba (changes requested):
@Frodigo We need to add information in docs how to handle this, or even that we have such a  custom configuration field.

Frodigo:
@bartoszherba documentation added.

PR #1256, opened 2022-07-13
fix: added limit for product gallery size

Frodigo:
### Before
When it was more than four images in the gallery, the layout of PDP was broken:
Screenshot 2022-07-13 at 15 28 11

### After
There can max four items in the product gallery

Screenshot 2022-07-13 at 15 29 01

**Note**: I know it's not a perfect solution, but because of Storefront UI constraints, I don't have a better idea for now. 
We need to wait until it is fixed in SFUI.

### DIFF
--- a/packages/theme/modules/catalog/product/getters/productGetters.ts
+++ b/packages/theme/modules/catalog/product/getters/productGetters.ts
@@ -89,7 +89,7 @@ export const getPrice = (product: ProductInterface): Price => {
   };
 };
 
-export const getGallery = (product: Product): MediaGalleryItem[] => {
+export const getGallery = (product: Product, maxGallerySize = 4): MediaGalleryItem[] => {
   const images = [];
 
   if (!product?.media_gallery && !product?.configurable_product_options_selection?.media_gallery) {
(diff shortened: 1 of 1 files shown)


packages/theme/modules/catalog/product/getters/productGetters.ts
```diff
@@ -90,6 +90,7 @@ export const getPrice = (product: ProductInterface): Price => {
 };
 
 export const getGallery = (product: Product): MediaGalleryItem[] => {
+  const MAX_GALLERY_SIZE = 4;
```
bartoszherba, line 93: Can we make it a getGallery param with a default value === 4?

Frodigo: Done

packages/theme/modules/catalog/product/getters/productGetters.ts
```diff
@@ -89,7 +89,7 @@ export const getPrice = (product: ProductInterface): Price => {
   };
 };
 
-export const getGallery = (product: Product): MediaGalleryItem[] => {
+export const getGallery = (product: Product, maxGallerySize = 4): MediaGalleryItem[] => {
```
bartoszherba, line 92:
```suggestion
export const getGallery = (product: Product, maxGallerySize: number = 4): MediaGalleryItem[] => {
```

Issue #1266, opened 2022-07-15
[Bug]: the selectedGallery from getGalley/productGetters will always return true on configurable products


bensinca:
### Expected Behavior

The `media_gallery` property is an array that as a first operand in a conditional operator returns `true` regardless, therefore, it should return a bool based on the length of the array.

If `media_gallery` is empty it should be expected to return false in order to return the `product.media_gallery` instead. 

### Actual Behavior

The conditional operator on configurable products will always return true.

### What version of Magento 2 integration are you using?

2.4.3

### What version of Node.js are you using?

16.15.0

### What browser (and version) are you using?

Chrome

### What operating system (and version) are you using?

macOS

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1267, opened 2022-07-15
Fix bug when media_gallery array is empty

bensinca:
In `getGallery` rather check the length of an array to get an accurate bool value than simple the array which will always return true.

## Related Issue
#1266

## Motivation and Context
If product type is `configurable_product` then check the length of the `media_gallery` array to get an accurate bool value, as an array will always return true.


Frodigo:
Hi @bensinca, thanks for your contribution. Could you change the base branch from the main to the develop branch?

bartoszherba (approved):

Frodigo (approved):

PR #1271, opened 2022-07-15
fix: fixed checkout issues

RetroModernDev:
Port of ##1270 to the develop


packages/theme/modules/checkout/pages/Checkout/ThankYou.vue
```diff
@@ -103,7 +103,7 @@ export default defineComponent({
     const route = useRoute();
 
     const companyDetails = ref({
-      name: 'Divante Headquarter',
+      name: 'Vue Storefront Headquarter',
```
bartoszherba, line 106: Can we add a dummy address? Or the actual address of VSF? Or just a placeholder? Nothing critical, just VSF is not in Wroclaw :)

bartoszherba (approved):

PR #1273, opened 2022-07-18
ci: simplify getting middleware url

sethidden:
the current middleware url getting logic can be moved to the already existing block

(still have to check if it works)


sethidden:
@Frodigo As a test deployed to enterprise env -  https://demo-magento2-enterprise.europe-west1.gcp.storefrontcloud.io/default/c/men/tops-men/jackets-men.html - it works so I assume middleware_url is passed to the built container fine

Frodigo (approved):

PR #1287, opened 2022-07-20
ci: refactor k6 action to add new env vars

sethidden:
Change K6 action so that it works by picking the environment rather than specifying the url. Our upcoming implementation of load tests will have three separate urls (magento instance, middleware url, frontend url). Specifying each url by hand/adding it to the url dropdown would be annoying

### DIFF
--- a/.github/workflows/deploy-vue-storefront-cloud.yml
+++ b/.github/workflows/deploy-vue-storefront-cloud.yml
@@ -17,47 +17,43 @@ jobs:
       deployment_id: ${{ steps.deployment.outputs.deployment_id }}
     steps:
       - name: Determine environment-specific variables
-        id: determine-environment
         shell: bash
         run: |
           REF=${{ github.ref }}
 
           if [ $REF = 'refs/heads/main' ]; then
-            ENVNAME='production'
-            ENVCODE='demo-magento2'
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-vue-storefront-cloud.yml
```diff
@@ -17,47 +17,43 @@ jobs:
...
-              echo 'unrecognized branch name'
-              exit 1
+            echo 'unrecognized branch name'
+            exit 1
           fi
 
-          echo ::set-output name=name::$ENVNAME
-          echo ::set-output name=code::$ENVCODE
-          echo ::set-output name=middleware-url::$MIDDLEWARE_URL
+          cat .github/workflows/public_env_info/$ENVINFO_FILE >> $GITHUB_ENV
```
sethidden, line 41: environment-related env vars are now kept in separate files. Had I not done that, env info would be duplicated in the k6 action and in our regular deploy to Farmer action. It'd be more difficult to maintain

bartoszherba (approved):

sethidden:
@bartoszherba this PR is targeting the main branch because workflows that are meant to be ran manually take source code from the repo's designated primary branch (main in our case)

Frodigo (approved):

PR #1294, opened 2022-07-21
fix: update vue dependencies

bartoszherba:
## Description
- lock dependencies on the last working version to avoid upgrade issues

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/package.json
+++ b/packages/theme/package.json
@@ -57,9 +57,12 @@
     "pinia": "^2.0.14",
     "pm2": "^5.2.0",
     "vee-validate": "^3.4.14",
-    "vue": "^2.6.14",
+    "vue": "2.6.14",
+    "vue-demi": "0.13.1",
     "vue-lazy-hydration": "^2.0.0-beta.4",
-    "vue-scrollto": "^2.20.0"
+    "vue-scrollto": "^2.20.0",
+    "vue-server-renderer": "2.6.14",
+    "vue-template-compiler": "2.6.14"
   },
   "devDependencies": {
     "@babel/core": "^7.18.6",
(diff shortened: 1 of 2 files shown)


packages/theme/package.json
```diff
@@ -57,9 +57,12 @@
...
     "pm2": "^5.2.0",
     "vee-validate": "^3.4.14",
-    "vue": "^2.6.14",
+    "vue": "2.6.14",
+    "vue-demi": "0.13.1",
     "vue-lazy-hydration": "^2.0.0-beta.4",
-    "vue-scrollto": "^2.20.0"
+    "vue-scrollto": "^2.20.0",
+    "vue-server-renderer": "2.6.14",
+    "vue-template-compiler": "2.6.14"
```
sethidden, line 65: Given this wasn't here before those are probably dependencies of `vue` itself, so if vue is 2.6.14 they should be 2.6.14 aswell.
I know they were in my yarn add magic command on Slack but they more fixed the yarn lock than being strictly necessary.

I'm accepting the PR either way

sethidden (approved):

packages/theme/package.json
sethidden: Actually deps of nuxt vue-renderer, so perhaps it'd be wiser to lock that version 
image

Frodigo: It didn't work witwithout this, because yarn installed vue-server-rendered@2.7.x

Frodigo (approved):

packages/theme/package.json
sethidden: Yes it probably didn't work without this because another package is installing the too-new version (in this case @nuxt/vue-renderer)

PR #1303, opened 2022-07-22
ci: stop using personal access token for deployment api

sethidden:
our approach to reporting deployment status changed so we no longer need
the personal access token workaround. In the past, it was needed because
the deploy-template.yml workflow was called by the parent
deploy-vue-storefront-cloud workflow, which is prevented by:
https://github.com/chrnorm/deployment-action/tree/3afecce166a0a2f92d3a9e7351a9908a72121c24#notes

using a personal access token is also not secure since if stolen it
could be used to push to the deployment status of every repo I have
access to (including vsf org).

### DIFF
--- a/.github/workflows/deploy-vue-storefront-cloud.yml
+++ b/.github/workflows/deploy-vue-storefront-cloud.yml
@@ -10,6 +10,10 @@ on:
 jobs:
   create-deployment:
     runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


sethidden:
The PAT also expires so this is just one less maintenance job to do.

bartoszherba (approved):

PR #1310, opened 2022-07-26
refactor: remove style duplication in my account route components

bartoszherba:
## Description
remove style duplication in my account route components

## Motivation and Context
Refactoring

## How Has This Been Tested?
-

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/modules/customer/pages/MyAccount/MyReviews.vue
+++ b/packages/theme/modules/customer/pages/MyAccount/MyReviews.vue
@@ -90,6 +90,8 @@ export default defineComponent({
 </script>
 
 <style lang="scss" scoped>
+@import '../styles/shared.scss';
+
 .no-reviews {
   &__title {
     margin: 0 0 var(--spacer-lg) 0;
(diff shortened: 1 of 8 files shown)


packages/theme/modules/customer/pages/MyAccount/MyReviews.vue
```diff
@@ -98,14 +100,6 @@ export default defineComponent({
...
 
-.tab-orphan {
-  @include for-mobile {
-    --tabs-title-display: none;
-    --tabs-content-padding: 0;
-    --tabs-conent-border-width: 0;
-  }
-}
-
 .message {
```
sethidden, line 103: You're importing shared.scss which contains the `.message` class, but on this line there's also a .message class. Doesn't this mess with the styles? Even if not, I think it'd be worth it to rename resolve this naming conflict

sethidden: The .message class from shared.scss and from here is also somewhat similar (only difference is margin, no?), would it be possible to streamline into one .message class?

sethidden (dismissed):

sethidden (approved):

PR #1322, opened 2022-08-02
chore: removed depracated composables packages from the repository

Frodigo:
## Description
Removed deprecated module from the monorepo. 

## Related Issue
M2-925

## Motivation and Context
Clean codebase, remove deprecated code


sethidden (approved):

tafelnl:
Could someone explain to me why these composables were deprecated and eventually removed from this package? I understand some of them have been moved to the `@vue-storefront/magento-theme` package. But where did `useProduct` go for example?

I think it's weird for a few reasons:
- [the docs](https://docs.vuestorefront.io/magento/composition/list-of-composables.html) still refer to the deprecated composables, so those are all 404s
- in the [integration template by the VSF-team](https://github.com/vuestorefront/ecommerce-integration-boilerplate/tree/main/packages/composables), the composables still live inside the `packages/composables` folder.
- all other integrations (e.g. salesforce commerce cloud and shopify) also have their composables live inside the `packages/composables` folder.

Why did you choose to do otherwise? And were do the current composables like `useProduct` live right now. And how could we contribute to them?

bartoszherba:
@tafelnl 
Composables package was removed and moved to the theme as it is a storefront package and belongs to the nuxt and the storefront context (https://github.com/vuestorefront/storefront-nuxt2-magento2) The `api-client` package and all server-side functions are in (https://github.com/vuestorefront/magento2).

All agnostic composables (without a named domain) are in `root/composables` in the storefront package. All composables that belong to a business domain like checkout, catalog, or customer, were moved to the related `modules/*` namespace. You can find `useProduct` in the `modules/catalog/product/composables`.

This `https://github.com/vuestorefront/ecommerce-integration-boilerplate` is an outdated repository, unmaintained for about 2 years, and will be archived this week.

tafelnl:
Thanks for your elaborate answer. I wasn't aware of the deprecation of that boilerplate. In the docs it's still mentioned: https://docs.vuestorefront.io/v2/integrate/integration-guide.html#project-structure

Are there updated docs available, or where did you get this information from?

tafelnl:
I did some digging, but I think this new structure is weird compared to the old one. The composables are now inside the boilerplate/starter instead of published through a npm package. This implies it would be fine if the developer would change the code inside them. Which I think should be discouraged. Also it would lead to unDRY code, because instead of the code being published and pulled through npm the composables are literally inside the developers codebase.

Also what if I don't care about the boilerplate/starter at all, but just want to start out fresh? Or if I'm currently using VSF Shopify but want to migrate to Magento? That would be a pain in the ass. While VSF tries to be platform/framework agnostic, so it largely defies the purpose.

Am I just completely misunderstanding some stuff here, or what's going on?

filrak:
Hey! 

Thanks for caring and sharing your opinion, we really appreciate that.

The reason why we moved composables to the theme was based on the feedback we were getting from our customers and users. Our composables enforce a certain way of interacting with Magento backend that turned out to be limiting for many of our customers. At first we were trying to make them more generic but we quickly realized that this is not the way.

Our current approach learns towards clean separation between business logic and Ui logic where business logic/integration logic (API client) resides on the server context and is often abstracted while UI logic (composables/components) is part of the theme that, we believe should be fully customizable as a boilerplate. This way we can achieve much better customizability on the UI layer.

Reg upgradability - we believe that the UI logic shouldn't be upgradable because the way how composables are structured would depend on how the UI is structured. Eg you could have just 1 composable for a Product Details Page that fetches product details, reviews and recommendations at once to decrease the number of network requests, you could have a separate composable for a quick view of a product in modal from Product Listing Page etc.

> Also what if I don't care about the boilerplate/starter at all, but just want to start out fresh? 

Ability to "start fresh" is also limited right now because composables package is tied to Nuxt context.

With a clear separation of UI and integration logic you'd be able to start fresh and use the middleware with ANY frontend framework, not only Nuxt. This is actually something we're currently working on and teased in [this post](https://vuestorefront.io/blog/removing-storefront-theme-from-magento-2-monorepo) and in the [January's issue](https://vuestorefront-8443671.hs-sites.com/vue-storefront-developer-newsletter-3-our-new-years-resolutions?ecid=ACsprvur7_1CV1Q--3Lw1JsFOh1siCPnUsZcMK0uRS1TknlxnFzQwgw0RDOoJn_nTg1XHUOym8Nm&utm_campaign=Community%20Newsletter&utm_medium=email&_hsmi=240839746&_hsenc=p2ANqtz--AnDxD294HR9Qy-9b-iohNfcI5Na9W3X4sMDAxcTnYRZtBggmaCJgNPQ1Nn96FEwp9C2KJUK_lrc6bP18iEWPmujdQHUMkGG7WT2w9puc1xk7xCVs&utm_content=240836890&utm_source=hs_email) of [developer newsletter](https://vuestorefront.io/developer-newsletter)

> Or if I'm currently using VSF Shopify but want to migrate to Magento? That would be a pain in the ass. While VSF tries to be platform/framework agnostic, so it largely defies the purpose.

This is a use case that we hoped to achieve with unified interface of composables and getters but we learned that this approach is very limiting for more complex cases and mostly suitable for small shops while VSF is a software primarily for larger ones with "exotic" use cases. Unification on the data later is very limiting, but we believe that the unification on workflows, architecture and tech stack is fulfilling the same goals while keeping the stack flexible enough to cover non-standard case.

tafelnl:
> Thanks for caring and sharing your opinion, we really appreciate that.

I'm happy you are open to discussion and opinions :+1: 

> With a clear separation of UI and integration logic you'd be able to start fresh and use the middleware with ANY frontend framework, not only Nuxt. This is actually something we're currently working on and teased in [this post](https://vuestorefront.io/blog/removing-storefront-theme-from-magento-2-monorepo) and in the [January's issue](https://vuestorefront-8443671.hs-sites.com/vue-storefront-developer-newsletter-3-our-new-years-resolutions?ecid=ACsprvur7_1CV1Q--3Lw1JsFOh1siCPnUsZcMK0uRS1TknlxnFzQwgw0RDOoJn_nTg1XHUOym8Nm&utm_campaign=Community%20Newsletter&utm_medium=email&_hsmi=240839746&_hsenc=p2ANqtz--AnDxD294HR9Qy-9b-iohNfcI5Na9W3X4sMDAxcTnYRZtBggmaCJgNPQ1Nn96FEwp9C2KJUK_lrc6bP18iEWPmujdQHUMkGG7WT2w9puc1xk7xCVs&utm_content=240836890&utm_source=hs_email) of [developer newsletter](https://vuestorefront.io/developer-newsletter)

This sounds awesome! Very exciting.

> Ability to "start fresh" is also limited right now because composables package is tied to Nuxt context.

Actually with "fresh", I meant more like, without all the theme stuff. So just a "fresh" nuxt installation. And then only install the composables. This would grant more flexibility for fully customized themes without using VSF UI at all. But I guess the framework-agnostic approach you mentioned solves that as well indeed!  

> Eg you could have just 1 composable for a Product Details Page that fetches product details, reviews and recommendations at once to decrease the number of network requests, you could have a separate composable for a quick view of a product in modal from Product Listing Page etc.

Wouldn't request batching solve this particular problem as well? That said this particular use-case maybe was just an example and you might have other good reasons to choose this route.

Though I still have trouble understanding exactly how and what your future approach will solve. Because wouldn't it be exactly the same as just implementing everything all by yourself? I mean, if it isn't commerce-agnostic in any way, what would be the benefit of using VSF magento over just using the GraphQL API of Magento directly? Will you be making the API-layer agnostic? Because that doesn't seem to be the case right now. As far as I understand VSF is mainly made agnostic by the use of its getters.

Also, if the unification of data is limited, because of exotic use cases, why build a VSF UI at all? I think the UI is where most of the magic will happen right? And if that UI cannot really be supplied with data in an agnostic way, why wouldn't the developer just build those components all by himself?

Maybe I'm thinking about this the wrong way, but even with the exotic use cases I have in mind, it would still be helpful to have a unified (and well documented in VSF) way to query data right? I'd like to think about VSF as some type of "ORM for ecommerce". I think your point about more complex use-cases like querying a product _with_ its reviews is a fair one. But I also think it could still be implemented in VSF in an agnostic manner that's still very flexible. One idea would be to support some type of object querying like prisma does. For example:

```
useQuery({
  query: {
    products: {
      reviews: true
    }
  }
})
```

This is complicated, I get that. But I think it would be doable.

Also stuff like adding a product to your cart, removing from your cart, updating the number of items, etc. etc. are most likely being used by even the most exotic use cases. It would make sense to abstract those right?

---

And I realize, as the saying goes, "Bachelors’ wives and maidens’ children are well taught.". So please do not get me wrong here. I understand that you guys probably thought very well about your decisions. I'm just mainly trying to understand them. And where I don't give my feedback as fairly as I could possibly do.

PR #1347, opened 2022-08-30
fix: m2-1017 logout doesn't work after reload of the MyAccount page

bartoszherba:
## Description
Logout should work always, also after page refresh during browsing "My Account".

## How Has This Been Tested?
Open Magento 2 e-commerce integration page.
2. Login using valid credentials.
3. Move the mouse to the "User Account" icon in the top menu bar and click it.
4. Refresh the page.
5. Move the mouse to "Logout" in the left column and click it.

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/modules/customer/pages/MyAccount/useSidebarLinkGroups.ts
+++ b/packages/theme/modules/customer/pages/MyAccount/useSidebarLinkGroups.ts
@@ -1,34 +1,39 @@
 import type { RawLocation } from 'vue-router';
+
 import { useRouter, useContext } from '@nuxtjs/composition-api';
 import { useUser } from '~/modules/customer/composables/useUser';
 import { useCart } from '~/modules/checkout/composables/useCart';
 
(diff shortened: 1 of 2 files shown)


packages/theme/modules/customer/pages/MyAccount/useSidebarLinkGroups.ts
```diff
@@ -38,26 +43,28 @@ export const useSidebarLinkGroups = () => {
...
         },
         {
+          id: 'my-reviews',
           label: 'My reviews',
           link: { name: 'customer-my-reviews' },
         },
         {
+          id: 'log-out',
           label: 'Log out',
-          listeners: {
```
bartoszherba, line 50: Unfortunately, if the component returns an object with a function inside, that function will be removed on the SRR, probably due to the serialization of the response. Only functions returned without containers are returned as expected.

Frodigo (approved):

PR #1351, opened 2022-08-31
fix: m2-1050 pods does not get up after deployment

bartoszherba:
## Description
- add ssrMiddlewareUrl to prevent make a call to unavailable external client

It's SSR <--> API implementation issue when it comes to deployment using a rolling-update strategy. During the deployment, the Kubernetes is probing SSR if it's already up. SSR is trying to call middleware using an external URL (market...[storefrontcloud. io](http://storefrontcloud.io/)) rather than localhost:3000/API, while the middleware isn't available publicly yet, because the SSR didn't respond with 200 in the first place. The workaround works because it allows the project to be deployed and make the API available, and later we fix the SSR to use the remote API address back again

## Types of changes

## Checklist:


Frodigo (changes requested):
Looks good to me, thanks! 

Can you update documentation and explain what ssrMiddlewareUrl variable is?

Frodigo (approved):

PR #1379, opened 2022-10-11
fix: product list page items on product page list page zero prices issues

bartoszherba:
## How Has This Been Tested?
1. Visit M2 integration app.
2. Clear cache.
3. Open any category page.
4. Open any product detail page.
5. Go back using native, browser back button.
6. Switch to different category.

or simply jump between Gear and Man categories.

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/modules/catalog/pages/category.vue
+++ b/packages/theme/modules/catalog/pages/category.vue
@@ -205,12 +205,12 @@ export default defineComponent({
     const { activeCategory, loadCategoryTree } = useTraverseCategory();
     const activeCategoryName = computed(() => activeCategory.value?.name ?? '');
 
+    const categoryUid = routeData.uid;
+
     const { fetch } = useFetch(async () => {
       if (!activeCategory.value) {
         await loadCategoryTree();
       }
-
-      const categoryUid = routeData.uid;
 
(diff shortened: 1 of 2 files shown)


packages/theme/modules/catalog/pages/category.vue
```diff
@@ -237,17 +237,27 @@ export default defineComponent({
...
       loadWishlist();
       const { getPricesBySku } = usePrice();
       if (products.value.length > 0) {
         const skus = products.value.map((item) => item.sku);
         const priceData = await getPricesBySku(skus, pagination.value.itemsPerPage);
-        products.value = products.value.map((product) => ({
-          ...product,
-          price_range: priceData.items.find((item) => item.sku === product.sku)?.price_range,
-        }));
+        products.value = products.value.map((product) => {
```
jeffpdotone, line 247: I believe mapping could be better if handled in the server-side instead of this kind.

But if it's really needed perhaps, we could extract them to a helper function or something similar.

bartoszherba: It can't be done on the server because it must bypass the cache including Redis or CDN, that is why it is executed in the `onMounted` hook otherwise, prices will be cached.

PR #1418, opened 2023-02-08
Ignore common URL parameters for paid advertising

davidwindell:
## Description
Url parameters such as `gclid` are common when using paid advertising, these currently cause category pages to 404 as the parameter gets passed to Magento as a search query.

## Motivation and Context
Prevents `gclid` and `fbclid` being passed on to Magento.

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/composables/useUiHelpers/index.ts
+++ b/packages/theme/composables/useUiHelpers/index.ts
@@ -4,7 +4,7 @@ import type { UseUiHelpersInterface } from '~/composables';
 import type { Params, QueryParams, FilterParams } from './Params';
 import type { FacetInterface } from '~/modules/catalog/category/types';
 
-const nonFilters = new Set(['page', 'sort', 'term', 'itemsPerPage']);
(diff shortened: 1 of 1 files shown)


bartoszherba:
@davidwindell Hello David, thank you for the pull request.
Currently, the list of parameters `const nonFilters = new Set(['page', 'sort', 'term', 'itemsPerPage']);` is used and required by the VSF to works properly. Additional parameters are not required and therefore cannot be a part of the core application. If a developer creates a module that requires or modify that list then it is responsibility of the module to change that. This list is potentially endless and keeping it as short as possible in the core application is the way to go.

davidwindell:
Do you have an alternative proposal for users of Google Ads etc to stop the
storefront breaking?

Right now if you are running ads to a category URL it will return a 404.

bartoszherba:
@davidwindell Yes, the proposition is to apply custom changes in a custom project as you did in this PR. Whoever is implementing that particular solution in his project can use this solution as a recommended approach. It can be described in the docs.

I am not saying that the solution is wrong. I say that the reason for this parameters is external for the VSF template and as such should be provided along with the external module.

davidwindell:
Understood :+1:

PR #1445, opened 2023-07-11
feat: update middleware to 3.2.0

michaelKurowski:
Update middleware to 3.2.0


michaelKurowski:
@sethidden Bartek said that something with the repo broke and there are no necessary secrets available

